PR #2, opened 2019-04-29
Add "fish" operators

TheMatten:
Based on https://github.com/isovector/polysemy/pull/24.


src/Polysemy/Fish.hs
```diff
@@ -0,0 +1,282 @@
...
+-- 'Sem' with list of multiple effects:
+--
+-- @
+-- foo :: 'Sem' \'['Polysemy.State.State' 'Int', 'Polysemy.Input.Input' 'String'] ()
+-- @
+--
+-- can be written as:
+--
+-- @
+-- foo :: \'['Polysemy.State.State' 'Int', 'Polysemy.Input.Input' 'String'] \@> ()
```
isovector: I'm not sure this is a good example, ideally people will never write a `Sem` like that.

isovector: Maybe show `State Int ': r` instead?

src/Polysemy/Fish.hs
```diff
@@ -0,0 +1,282 @@
...
+--
+-- can be written simply as:
+--
+-- @
+-- foo :: ConsoleIO:r \@> () -> 'IO' >\@ r \@> ()
+-- @
+--
+-- and will be automatically lifted and put into list.
+--
+-- __NOTE:__ For this operator to work, you have to provide types of monad and
```
isovector: "you need to provide an explicit `* -> *` kind signature on the `m` parameter"

isovector: I can look at convincing the plugin to provide these kind sigs too; I don't think it would be very hard.

src/Polysemy/Fish.hs
```diff
@@ -0,0 +1,282 @@
...
+--
+-- can be written as:
+--
+-- @
+-- foo :: 'String' -> 'Int' -> \'['Polysemy.State.State' 'String', 'Polysemy.Input.Input' 'Int'] >\@> ()
+-- @
+--
+-- For more examples on how to specify member effects, see ('>@').
+--
+-- __NOTE:__ For this operator to work, you have to provide types of monad and
```
isovector: Same change

src/Polysemy/Fish.hs
```diff
@@ -0,0 +1,282 @@
...
+--
+-- __NOTE:__ For this constraint to work, you have to provide types of monad
+-- and contained value in your effect:
+--
+-- @
+-- data MyEffect x y (m :: * -> *) (a :: *) where ...
+-- @
+--
+-- but this requirement may be removed in the future.
+type Means es r = Members (ToEffList "Means" es) r
```
isovector: I'm not sure it's worth exposing this --- what do you think is the benefit?

src/Polysemy/Fish.hs
TheMatten: Ok, will do.

src/Polysemy/Fish.hs
TheMatten: Not really, I just added it for consistency. Will be removed.

TheMatten:
After I add some example with interpreter, it should be ready for merging :smile:

isovector:
Fantastic, thanks!!!

PR #3, opened 2019-04-30
Ported Random from polysemy, added RandomFu.  

adamConnerSax:
Also: Updated package.yaml and regenerated polysemy-zoo.cabal. Added some things to gitignore for cabal v2-style builds and emacs TAGS files.

Questions:  
1. Are we comfortable with the orphan instance of "MonadRandom"?  I think it will be useful but it's somewhat against the spirit of polysemy's mtl-free-ness and it's an Orphan.

2. I don't see an easy way to implement these as the same effect and then interpret but I didn't look that hard.  I can imaging Random being implemented in terms of RandomFu but not the other way around.  And that's a commitment to a less standard external dependency.

3.  It would be possible, but, I think, a bit more work to write the equivalent of ```runRandom``` for RandomFu.  random-fu has instances of things in terms of ```State StdGen``` but those are built around the mtl ```State``` and they'd have to be re-written, which would mean more orphan instances. I think.

4. Should there be tests?


src/Polysemy/RandomFu.hs
```diff
@@ -0,0 +1,110 @@
...
+-- | Run a 'Random` effect by using the default "random-fu" 'IO' source
+runRandomIO
+  :: forall r a
+   . MonadIO (Sem r)
+  => Sem (Random ': r) a
+  -> Sem r a
+runRandomIO = interpret f
+ where
+  f :: forall m x . (Random m x -> Sem r x)
+  f r = case r of
```
isovector: Can this be inlined with `LambdaCase` and/or `BlockArguments`?

src/Polysemy/RandomFu.hs
```diff
@@ -0,0 +1,110 @@
...
+  :: MonadIO (Sem r)
+  => R.PureMT
+  -> Sem (Random ': r) a
+  -> Sem r a
+runRandomIOPureMT source re =
+  liftIO (newIORef source) >>= flip runRandomSource re
+{-# INLINE runRandomIOPureMT #-}
+
+-- | Orphan instance of 'R.MonadRandom'
+$(R.monadRandom [d|
```
isovector: Why is this template haskell? 

I'd prefer not introducing an orphan here, but I could be talked into it if I knew why it's necessary.

src/Polysemy/RandomFu.hs
```diff
@@ -0,0 +1,110 @@
...
+
+------------------------------------------------------------------------------
+-- | Run in 'IO', using the given 'R.PureMT' source, stored in an 'IORef'
+runRandomIOPureMT
+  :: MonadIO (Sem r)
+  => R.PureMT
+  -> Sem (Random ': r) a
+  -> Sem r a
+runRandomIOPureMT source re =
+  liftIO (newIORef source) >>= flip runRandomSource re
```
isovector: To confirm my understanding, an `IORef PureMT` is a `RandomSource`?

src/Polysemy/RandomFu.hs
```diff
@@ -0,0 +1,110 @@
...
+import qualified System.Random                 as SR
+import           Control.Monad.IO.Class         ( MonadIO(..) )
+
+------------------------------------------------------------------------------
+{- | An effect capable of sampling from a "random-fu" RVar or generating a
+single random-variate of any type, @t@ with a
+@Data.Random.Prim t@ constructor, currently one of @Word8@, @Word16@,
+@Word32@, @Word64@, @Double@ or N-byte integer.
+-}
+data Random m r where
```
isovector: RandomFu

polysemy-zoo.cabal
```diff
@@ -28,6 +28,8 @@ source-repository head
 library
   exposed-modules:
       Polysemy.KVStore
+      Polysemy.Random
```
isovector: I'd prefer if we move this in a separate PR

src/Polysemy/RandomFu.hs
adamConnerSax: Yes.  Random-fu is a little confusing to me.  There are sources ```s```, where ```RandomSource m s``` means that ```s``` can be "sampled" in ```m```.  There is also ```MonadRandom m``` which sort of implies that there exists a source such that it can be sampled in ```m```.  Users, in the sense of functions that would like to be able to sample random-variables, don't need to know about the source.  But when you run the thing, you do.  I think all of this is just to generalize over ```StdGen``` and ```PureMT``` and possible future generators.  But I'm not sure.

src/Polysemy/RandomFu.hs
adamConnerSax: I'm happy to make this change but it's hard to imagine circumstances where a user would want both and the name clashes would be an issue.

polysemy-zoo.cabal
adamConnerSax: Sure.  I'll take it out.  I thought it might facilitate the conversation about trying to unify them

isovector: Gotcha. I'm not sure it's worth spending time on --- we can let someone else figure out how to put them together if they care enough.

src/Polysemy/RandomFu.hs
adamConnerSax: The TH generates several more functions in the instance and the random-fu docs say one should do it this way since the implementation of those might change.
I have a bunch of data-science stuff that uses random-fu via the "MonadRandom" constraint.  So this makes that painless.  There might be other users like that.
I think this might be a general tension between thinking of effects as an app-level thing vs. as a specific utility.  I'm using it more in the latter way, where the effects may need to interact with existing monads/mtl constraints.  So I'm trying to make that easy where the costs are low.

src/Polysemy/RandomFu.hs
isovector: Yeah, I hear you. I think it's a good idea for a few reasons, though:

1. Reinforce the idea that the effect `E` comes from module `Polysemy.E`
2. Unique names let you disambiguate effects at a glance, rather than needing to look at which `Random` effect is being imported. Especially since they have very different APIs.

src/Polysemy/RandomFu.hs
isovector: Random fu is confusing to everyone :)

Is the `MonadRandom` bit necessary?

isovector:
To answer your questions: 

I'd love a pure `runRandomFu` function. Rather than introducing the orphan, you can almost certainly get it by reinterpreting `RandomFu` into `Lift (Control.Monad.State whateverthingyouneed)`, and then handling that thing explicitly.

Tests would be good -- if only to show how to use the thing.

adamConnerSax:
Fixed the simple stuff.  Renaming, LambdaCase, removing Random.  I'll have to think more about the MonadRandom/instance stuff.  So that'll have to wait a bit...

isovector:
> What do you mean in the above by "get it by"? What would I do with a function that has a MonadRandom constraint and which I would like to run in a Member RandomFu r => Sem r without the instance I provided?

There are `Monad m => MonadRandom (StateT StdGen m)` instances; so I don't see why we couldn't reinterpret a `Sem (RandomFu ': r) a` into a `Sem (State StdGen ': r) a` into a `StateT StdGen (Sem r) a` into `Sem r (StdGen, a)`.

The `Sem (State s ': r) a -> StateT s (Sem r) a`  transformation might be tricky, but I don't see any reason it couldn't be done.

isovector:
[`hoistStateIntoStateT`](https://github.com/isovector/polysemy/commit/bfaccf13e8a07ec37fd728624b15d89acc1f57a2#diff-fd3498976a9f73aa91ed6fd247ce00efR98)

adamConnerSax:
Nice.  That would've taken me...a while.

adamConnerSax:
So.  I get the runRandomFu thing and, once there is a released version of polysemy with the ```hoistStateIntoStateT```, I'll fix up the PR.

But I still don't get what your mean about the instance/constraint thing.  You may need to explain it to me like I am a (Haskell-knowing) five-year-old.
Suppose I have a function:
```haskell
doSomethingRandom :: MonadRandom m => m a
```
and I want to be able to call that function inside 
```haskell
myEffectfulFunction :: Member RandomFu r => (a -> b) -> Sem r b
myEffectfulFunction f = do
   a <- ?? doSomethingRandom
   return $ f a
```
either by having 
```instance Member RandomFu r => MonadRandom (Sem r)``` or what?  I guess a function of the form
```liftRandomFu :: (Member RandomFu r, MonadRandom m) => m a -> Sem r a``` ??
I don't quite see how hoisting to StateT gives us the former.  Am I missing something? 
And the latter, a function which discharges the constraint in the presence of ```Member RandomFu r ``` certainly seems possible, but also sub-optimal for ease of use.  Is that what you have in mind?  Sort of like ```liftIO```?  I guess I can see how that might be the right balance when trying to interoperate with mtl but not become shot through with it.

**Comment by isovector** [comment]
I was under the impression that the `MonadRandom` bit was necessary to run *any* of the `random-fu` stuff. If it's actually just for convenience of interoping with code that already uses `MonadRandom`, then yes, I agree, let's just burn the bridge and not support any `MonadRandom` stuff.

**Comment by adamConnerSax** [comment]
Hmm.  But I still want the interop! I have some data-science stuff that has that signature.  Which I guess I could change but there's a lot more mtl style code out there and it might be worth creating a path.  Let me think about the constraint-discharging version.  That wouldn't infect the lib with an orphan instance but would provide a way for interop to be relatively painless.

**Comment by isovector** [comment]
Hiya, what's the status of this?

**Comment by adamConnerSax** [comment]
Hiya!

I guess I was waiting for a release with hoistStateIntoStateT and then I would try to move forward.  Has that happened?  I get confused trying to work against unreleased versions...

In the meantime I have been using Polysemy a bunch in knit-haskell and doing some fun Bayesian inference on voter preference using US House Election Results.  And discovering a couple of usability things:

1.  (Without the plugin) Many errors become Ambiguous type errors.  It's a little like TypeApplications.  Any small thing breaks the inference and then the first error you get is always Polysemy.

2.  Trying to add the plugin adds a lower bound on containers of 0.6.0 (not sure why.  Maybe the "installed" ghc version?) and that makes it unusable for me because knit-haskell has a "plots" dependency with an < 0.6.0 bound on containers. Sigh.

Anyway, happy to get back to it.  Should I?

Adam

**Comment by isovector** [comment]
1. Try disabling the `custom-errors` flag --- it'll stop the aggressive polysemy errors. It's an unfortunate side-effect of GHC's error machinery. Maybe I'll investigate generating those errors from the plugin, where it'd be much easier to not clobber the usual Haskell errors.
2. Sorry about that, I needed to add explicit bounds for getting the package into stackage, and didn't feel like playing around to get a real set of bounds. If you can send a PR changing the `package.yaml` files for a better `containers` lower-bound I'll happy merge it and push a release ASAP.

**Comment by isovector** [comment]
As for `hoistStateT`, I think we came to the conclusion that it won't actually help with the `MonadRandom` instance.

**Comment by adamConnerSax** [comment]
I'll get back to it then!  For now, I can work around the containers issue with an "allow-newer" in my cabal.project.  Plots seems to compile fine with containers-0.6.

**Comment by adamConnerSax** [comment]
I fiddled with this for a bit and so far the only thing I can do is

```haskell
absorbMonadRandom
  :: Member RandomFu r => (forall m . R.MonadRandom m => m a) -> Sem r a
```
(just via an instance on a newtype wrapper)
which gets us some of the way but not as far as I'd like.  But I don't think we can do the more general
```haskell
absorbMonadRandom :: (MonadRandom m, Member RandomFu r) => m a -> Sem r a
```
because we can't the ```a``` out of that monad, right?  The case above works because of the Rank2-ness of the 'm' so the ```absorbMonadRandom``` function gets to choose to implement ```m``` as the wrapped ```Sem``` with the instance, thus discharging the constraint.  But the more I think about it, the less clear I get on how we could do more, especially without violating the spirit of Polysemy--to not prematurely foreclose on our possible interpretations.

I don't know how useful this thing will be.  I'm going to try and find some places to use it and see if it's helpful .

And "absorb"?  I don't know.  It's not quite lift or hoist or raise, right?

Once I use it a bit, I'll try to add an example that makes it clearer how to use it and random-fu altogether.  But that might need to wait until next week.

isovector:
I _really_ like the type of `absorbMonadRandom`. I think it's an elegant solution to the mtl interop problem. Maybe we could go further with constraint kinds:

```haskell
class CanonicalEffect (c :: (* -> *) -> Constraint) (e :: (* -> *) -> * -> *) where
  absorb :: Member e r => (forall m. c m => m a) -> Sem r a

instance CanonicalEffect MonadRandom Random where
instance CanonicalEffect (MonadReader r) (Reader r) where
```

etc., although I'm not sure how nicely it would play if you have multiple mtl constraints you want to absorb simultaneously.

adamConnerSax:
Thanks!

Yeah.  But maybe single effects are enough?  If you start having multiple effects you should re-write in polysemy or use polysemy on top or on the bottom of the stack, maybe?  Let me make some examples for this one and then think a bit about the generalized version of "absorb". And multiple effects.  

Would that class need a fundep (or an associated type family)?  How would the compiler know what ```e``` you had in mind at the ```absorb``` call-site?  Also, I wonder if all the absorbers are canonical?  Is there ever a case where we might choose different effects to absorb the same mtl-constraint?  For instance, if we could interpret random-fu via System.Random then we might want to absorb the ```MonadRandom``` constraint via ```Random``` instead of ```RandomFu```.  But maybe that's a bad example because the random-effects are specifically weird.  

But the question stands.  I'm not sure this generalization should be a typeclass though we can certainly generalize the signatures and add some helpers so that each specific absorber is more compact and uniform.

And I suppose you could write multi-effect absorbers. Not in the polysemy libraries, but for a specific application it might make sense as a way to use legacy mtl-code.  So maybe the "absorb" kit should be built to make those easy to build when you want them?  I'll think about that as well.

adamConnerSax:
See https://github.com/isovector/polysemy-zoo/pull/8 for CanonicalEffect.  I'll clean up this one once you decide how you want that to work...

PR #6, opened 2019-05-23
Add nightly builds to travis

isovector:
Let's make sure CI will yell at us before the nice stack people do. https://github.com/commercialhaskell/stackage/issues/4572


isovector:
So this should use a custom stack.yaml that doesn't specify the extra deps of `polysemy` and `polysemy-plugin`, but that's a task for another day.

PR #12, opened 2019-05-30
Fixed exports, of member functions and Data.Reflection.Reifies

adamConnerSax:
Sorry about this.  Was preparing to add tests for and re-submit the RandomFu stuff and ran into the fact that I forgot to export some necessary things for making absorbers in other modules using ```Polysemy.MTL```
I think that's fixed here.  And I've added the labor-saving export of ```Data.Reflection.Reifies```

### DIFF
--- a/src/Polysemy/MTL.hs
+++ b/src/Polysemy/MTL.hs
@@ -30,6 +30,8 @@ module Polysemy.MTL
   , Reifies
   , (:-)(Sub)
   , Dict(Dict)
+  , reflect
+  , Proxy (Proxy)
   )
 where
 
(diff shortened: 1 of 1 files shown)


adamConnerSax:
Actually.  Hold up.  A couple more incoming.  We should not name the ```Dict``` in ```ReifiableConstraint1``` "Dict" because of the collision with Constraints.Dict.  So I'l make it ```Dict1``` as it was.  And there needs to be an export of the constructor of ```ConstrainedAction``` or TH can't use it, which breaks RandomFu.  And I think I'll re-export ```Sub``` and ```Dict``` from constraints to save that dependence for users as well.
Working...

adamConnerSax:
Sigh.  There were a couple more.  Now I can implement the RandomFu absorber with no additional exports.
You can merge this now, or wait until I've added some RandomFu tests to make sure that doesn't catch something else in MTL somehow.

isovector:
Let me know when this is good to go!

adamConnerSax:
I think it's good to go.  I've got one working test in the RandomFu branch with this version of MTL.  So I think this has all the exports we need.  
I'm going to add more RandomFu tests before doing a PR for that one.

isovector:
Thanks!

1atAlcone (approved):

PR #13, opened 2019-05-31
Add RandomFu

adamConnerSax:
This seems in good enough shape to review.  I've added the effect as before.  Added the absorber using the code in MTL.  Added a couple of tests, though I'm open to more interesting or useful examples if anyone has ideas!

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -27,6 +27,9 @@ dependencies:
 - containers
 - reflection
 - constraints
+- random
+- random-fu
+- random-source
 
 ghc-options:
   - -fplugin=Polysemy.Plugin
--- a/package.yaml
+++ b/package.yaml
@@ -27,7 +27,6 @@ dependencies:
 - containers
 - reflection
 - constraints
-- random
 - random-fu
 - random-source
 
--- a/src/Polysemy/RandomFu.hs
+++ b/src/Polysemy/RandomFu.hs
@@ -1,0 +1,128 @@
+{-# LANGUAGE TemplateHaskell            #-}
+{-# LANGUAGE UndecidableInstances       #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-|
+Module      : PolySemy.RandomFu
+Description : Polysemy random-fu effect
+
+Polysemy "random-fu" effect.
+This can be run in a few ways:
+1. Directly in 'IO'
+2. Using any 'Data.Random.RandomSource' from "random-fu"
+3. In 'IO', using a given 'Data.Random.Source.PureMT' source. ('IO' is used to put the source in an 'IORef')
+
+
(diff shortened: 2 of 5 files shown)


src/Polysemy/RandomFu.hs
```diff
@@ -0,0 +1,133 @@
...
+  :: MonadIO (Sem r)
+  => R.PureMT
+  -> Sem (RandomFu ': r) a
+  -> Sem r a
+runRandomIOPureMT source re =
+  liftIO (newIORef source) >>= flip runRandomSource re
+{-# INLINEABLE runRandomIOPureMT #-}
+
+------------------------------------------------------------------------------
+absorbMonadRandom
```
isovector: Haddock for this please; even if it's just to point towards `absorb`.

test/RandomFuSpec.hs
```diff
@@ -0,0 +1,59 @@
...
+    >>= (`shouldBe` True)
+
+  it
+      "Should produce two distinct sets of psuedo-random Ints (absorber version)."
+    $   (runM . runRandomIO $ absorbMonadRandom $ do
+          a <- getRandomIntsMR 5
+          b <- getRandomIntsMR 5
+          return (a /= b)
+        )
+    >>= (`shouldBe` True)
```
isovector: I think this would be cleaner if you bound the monadic action above into a variable. Then you woudn't need the parens around the do block, we could get rid of the infix notation for `shouldBe`, and it would be easier to parse which monad that thing is running in at a glance.

src/Polysemy/RandomFu.hs
adamConnerSax: Yep.  Done.

test/RandomFuSpec.hs
adamConnerSax: Happy to bind to a variable.  Left the infix since I actually think it's clearer here.  How would you rather format it?

isovector: I'd suggest:

```
result <- runM ... .. $ do
  ...
result `shouldBe` True
```

instead of the `>>=` stuff

adamConnerSax: Fair enough.  Done, at least up to brittany formatting weirdness.

package.yaml
```diff
@@ -27,6 +27,9 @@ dependencies:
 - containers
 - reflection
 - constraints
+- random
```
isovector: I feel like this might be a leftover from the old PR when we also wanted to port `Random` from `polysemy`? But I don't know anything about random-fu so maybe this is a stupid comment and everything is as it should be!

isovector:
Looks great! Happy to merge when you give me the :+1:

package.yaml
adamConnerSax: Nice catch!  Removed. Good to merge!

Issue #16, opened 2019-06-02
RandomFu is a huge dependency


isovector:
Dang its _real_ big. Compilation times shot up drastically upon including it. I think we should think about moving it out to a separate package (`polysemy-random-fu` maybe?).

![wreq](https://user-images.githubusercontent.com/307223/58756767-f7607780-84cc-11e9-962e-2c9ec024524e.png)

isovector:
Before: ![wreq2](https://user-images.githubusercontent.com/307223/58770583-cbe99580-857d-11e9-9b92-279efe43329b.png)

adamConnerSax:
Separate package is fine! Or you can just drop it. I still have it in knit-haskell (hadn’t upgraded yet!) so I’m personally indifferent at the moment.

isovector:
@adamConnerSax would you want to be the maintainer of such a package? I'd prefer it be you instead of me, since I don't know anything about randomfu!

adamConnerSax:
Sure! You care if it’s on stackage? I’ve never dealt with that.

isovector:
I don't care! But putting it on stackage turns out to be super easy:
https://github.com/commercialhaskell/stackage/pull/4562/files

adamConnerSax:
I’ve seen that! Does it only work if you use hpack or stack? I’m sort of all cabal but willing to learn hpack...
You want me to just make a package with just RandomFu and publish or will you do it, leave in this repo and make me the maintainer?

adamConnerSax:
Bump.  Just making sure you're not waiting on me for something...
I'm happy to take over whatever when you want me to or to leave things as they are or whatever.

isovector:
Thanks for the reminder --- totally forgot about this. I'd prefer if you are solely responsible for the package:

1. Moving it into its own repo
2. Publishing it on hackage

if you're cool with that. Afterwards I'd be happy to link to your package from this one!

adamConnerSax:
Sure.  Might not happen until next week.  
Same namespacing, that is, call the module "Polysemy.RandomFu"?  Or do you want something different when they are external?
Maybe you should add an "EffectsPackages.md" file here with links?  Or a section at the bottom of the readme?

isovector:
The module name is fine! The dream is that no effects are more privileged
than any others. Good idea about the markdown file; will do after the
package is published.

isovector:
[...]

PR #20, opened 2019-06-08
Added CanonicalEffect instance for MonadRandom in RandomFu

adamConnerSax:
I forgot this before, the instance of the ```CanonicalEffect``` type-family for MonadRandom.  It's not urgent, but it ought to be in there.

### DIFF
--- a/polysemy-zoo.cabal
+++ b/polysemy-zoo.cabal
@@ -1,6 +1,6 @@
 cabal-version: 1.12
 
--- This file has been generated from package.yaml by hpack version 0.31.1.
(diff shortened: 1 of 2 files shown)


isovector:
Thanks for the PR! As a heads up, I noticed my compile times drastically spike after the randomfu stuff, and filed #18 about it. I'd like to move the random-fu effect out of this library, but wanted to talk to you about where we should put it!

adamConnerSax:
Yep. Responded there (which was https://github.com/isovector/polysemy-zoo/issues/16)

isovector:
Closing this due to the discussion in #16.

PR #25, opened 2019-06-15
Add hspec discover

adamConnerSax:
TL;DR: Adds ```hspec-discover``` as a build-tool dependency in ```package.yaml``` and the corresponding new cabal file.

The test suite won't build without hspec-discover being installed.  This bit me two ways:

- The first time I tried testing the zoo, I didn't have hspec-discover installed and so the tests wouldn't build.

- haskell-ci generated travis.yml wasn't working for RandomFu until I added hspec-discover as a build-tool dependency in the test section.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -62,6 +62,8 @@ tests:
     - -threaded
     - -rtsopts
     - -with-rtsopts=-N
+    build-tools:
+    - hspec-discover  >= 2.0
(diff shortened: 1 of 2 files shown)


isovector:
Awesome, thanks! This just works on stack, so it's nice to know my experience isn't universal. I'll make the same change in polysemy.

Issue #24, opened 2019-06-14
MTL absorbers are complex when they have superclasses


adamConnerSax:
This may be simpler than I think but I started to try and write the CanonicalEffect code for my PandocMonad class. ```PandocMonad``` has a superclass constraint of ```MonadError PandocError m``` that is, somewhere in Pandoc there is
```class (MonadError PandocError m) => PandocMonad m where ...```
and that seems to cause some weirdness:  

-The ```ReifiedConstraint1``` instance wouldn't compile, complaining about a functional 
dependency.

- I'm not sure how to write the superclass instance on ```ConstrainedAction PandocMonad``` itself, which I think looks like:
```haskell
instance (Monad m
         , P.Reifies s' (P.Dict1 (MonadError PA.PandocError) m)) => MonadError PA.PandocError (P.ConstrainedAction PA.PandocMonad m s') where
  throwError e = P.ConstrainedAction $ throwError_ (P.reflect $ P.Proxy @s') e
  catchError x f = P.ConstrainedAction
    $ catchError_ (P.reflect $ P.Proxy @s') (P.action x) (P.action . f) 
```
but ```throwError_``` and ```catchError_``` are not in scope.  Might just be an export issue?  But also feels like I'm just not seeing this clearly.

I haven't dug in yet so this is sort of preliminary but I wanted to signpost it here so I/we remember to document this and however we solve it, if it warrants explanation.

isovector:
Maybe `reflection` can be of some inspiration:

https://github.com/ekmett/reflection/blob/49d1d6ad2355f4cb2061bff0c744feb8841cd6e1/examples/Constraints.hs#L75-L83

adamConnerSax:
Okay.  Figured that out, I think.  Harder than the reflection example but that's likely because my brain frizzes a bit every time I see "Sub Dict" and it takes me a while to uncross the wires.

Anyway, solved it by (for the sake of documenting the process):
1. Adding ```throwError_``` and ```catchError_``` to the ```Dict1``` associated data type
2. Using them to create the instance above.
3. Filling the canonical dictionary bit of this with ```throw``` and ```catch``` from Polysemy.Error.

Which all makes sense.

But now life is full of Oprhan instances!  GHC says:
```haskell
Orphan instance: instance P.ReifiableConstraint1 PA.PandocMonad
    To avoid this
        move the instance declaration to the module of the class or of the type, or
        wrap the type with a newtype and declare the instance on the new type.
    |
[...]

PR #26, opened 2019-06-17
Non type class absorbers

adamConnerSax:
This is what I have in mind.  The code for each absorber is simpler and the typeclass stuff is all gone, since I think all of its cleverness was only useful for the absorbers in the MTL module.  This way we are setting a good example in the zoo by doing each one separately and exporting only the constraint-monomorphic absorber. And we are setting a good naming example for people to follow if they write their own absorbers for others to use.

Tests all pass unchanged.  The only change there was the name of the test file and the imports.  Which is reassuring.

That being said, it's a breaking change.

I'm going to re-write RandomFu's absorber this way and my PandocMonad absorber just to make sure it all works as simply as before, etc.


adamConnerSax:
Also, I'm not sure why it has so many commits.  I thought I had merged the upstream stuff before doing this work.  Is that an issue?  Is there something I can do on my end to fix it?

adamConnerSax:
One more question:  We don't need to export "using" from Polysemy.ConstraintAbsorber.  Should I drop it since ```absorbWithSem``` is the version we want people to use since it has the right signature and simpler order of type-variables for use in making absorbers?

src/Polysemy/ConstraintAbsorber.hs
```diff
@@ -0,0 +1,62 @@
+{-# LANGUAGE AllowAmbiguousTypes         #-}
+{-# LANGUAGE ConstraintKinds             #-}
+{-# LANGUAGE FlexibleInstances           #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving  #-}
+{-# LANGUAGE MultiParamTypeClasses       #-}
+{-# LANGUAGE ScopedTypeVariables         #-}
+{-# LANGUAGE TypeFamilies                #-}
+{-# LANGUAGE UndecidableInstances        #-}
+{-# LANGUAGE UndecidableSuperClasses     #-}
+{-# OPTIONS_GHC -fplugin=Polysemy.Plugin #-}
```
isovector: This is already specified in the cabal file, so unnecessary here.

src/Polysemy/ConstraintAbsorber.hs
```diff
@@ -0,0 +1,62 @@
+{-# LANGUAGE AllowAmbiguousTypes         #-}
+{-# LANGUAGE ConstraintKinds             #-}
+{-# LANGUAGE FlexibleInstances           #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving  #-}
+{-# LANGUAGE MultiParamTypeClasses       #-}
+{-# LANGUAGE ScopedTypeVariables         #-}
```
isovector: STV and TypeFamilies are already enabled on a project-wide basis. Please remove from here and other files.

src/Polysemy/ConstraintAbsorber.hs
```diff
@@ -0,0 +1,62 @@
...
+  (C.unsafeCoerceConstraint :: ((p (x m s) :- p m))) i
+{-# INLINEABLE using #-}
+  
+-- | Specialize using for the case when m ~ Sem r
+-- and rearrange type arguments for simpler typeapplication
+-- in use.
+absorbWithSem :: forall p (x :: (Type -> Type) -> Type -> Type ->Type) d r a. d
+              -> (forall s. R.Reifies s d :- p (x (Sem r) s))
+              -> (p (Sem r) => Sem r a) -> Sem r a
+absorbWithSem = using @d @x
```
isovector: Any reason we can't just inline the definition of `using` here, and then drop `using` entirely?

src/Polysemy/ConstraintAbsorber.hs
```diff
@@ -0,0 +1,62 @@
...
+  -> m a
+using d i m =
+  R.reify d $ \(_ :: Proxy (s :: Type)) -> m \\ C.trans
+  (C.unsafeCoerceConstraint :: ((p (x m s) :- p m))) i
+{-# INLINEABLE using #-}
+  
+-- | Specialize using for the case when m ~ Sem r
+-- and rearrange type arguments for simpler typeapplication
+-- in use.
+absorbWithSem :: forall p (x :: (Type -> Type) -> Type -> Type ->Type) d r a. d
```
isovector: Nit: please style types like this:

```haskell
foo
    :: forall (a :: k1)
              (b :: k2)
     . Monad m
    => d
```

src/Polysemy/ConstraintAbsorber/MonadError.hs
```diff
@@ -0,0 +1,52 @@
...
+    throwError_ :: forall a. e -> m a
+  , catchError_ :: forall a. m a -> (e -> m a) -> m a
+  }
+
+-- | Wrapper for a monadic action with phantom
+-- type parameter for reflection.
+-- Locally defined so that the instance we are going
+-- to build with reflection must be coherent, that is
+-- there cannot be orphans.
+newtype Action m s' a = Action { action :: m a } deriving (Functor, Applicative, Monad)
```
isovector: It might be clearer to name this `Orphan`.

src/Polysemy/ConstraintAbsorber/MonadError.hs
```diff
@@ -0,0 +1,52 @@
...
+  => (S.MonadError e (Sem r) => Sem r a) -> Sem r a
+absorbError = absorbWithSem @(S.MonadError _) @Action
+  (ErrorDict throw catch)
+  (Sub Dict) 
+{-# INLINEABLE absorbError #-}
+
+-- | A dictionary of the functions we need to supply
+-- to make an instance of Error
+data ErrorDict e m = ErrorDict
+  {
```
isovector: Nit: here and elsewhere, prefer to put the first element on the same line as the opener, eg:

```haskell
data Foo = Foo
  { throwError_ :: ..
```

likewise for module export lists

src/Polysemy/ConstraintAbsorber/MonadError.hs
```diff
@@ -0,0 +1,52 @@
...
+-- there cannot be orphans.
+newtype Action m s' a = Action { action :: m a } deriving (Functor, Applicative, Monad)
+
+-- | Given a reifiable mtl Error dictionary,
+-- we can make an instance of @MonadError@ for the action
+-- wrapped in @Action@.
+instance ( Monad m
+         , Reifies s' (ErrorDict e m)
+         ) => S.MonadError e (Action m s') where
+  throwError e = Action $ throwError_ (reflect $ Proxy @s') e
```
isovector: I just realized I have no idea how this works. How come giving an instance for `Action` gives us an instance for `Sem`? Do we unsafe coerce the dict somewhere? Is this how it used to work too, and I've just been cargo-culting all along?

src/Polysemy/ConstraintAbsorber/MonadWriter.hs
```diff
@@ -0,0 +1,66 @@
...
+  let semTell = tell
+      semListen :: Member (Writer w) r => Sem r b -> Sem r (b, w) 
+      semListen = fmap (\(x,y) -> (y,x)) . listen @w
+      semPass ::  Member (Writer w) r => Sem r (b, w -> w) -> Sem r b
+      semPass x = do
+        (w, (a, f)) <- listen x
+        censor f (tell w >> pure a)
+  in absorbWithSem @(S.MonadWriter _) @Action 
+     (WriterDict semTell semListen semPass)
+     (Sub Dict)
```
isovector: Is it possible to factor this `Sub Dict` into `absorbWithSem` or does it break the constraint machinery somehow?

isovector (approved):
I love this --- it's so much prettier than the original stuff. Great initiative! Left some stylistic points and some questions, but happy to merge when you're ready.

isovector:
As for why the extra commits, I'm not sure. Do you have an upstream remote set on your local repo? I think it might be using the wrong merge strategy, but it's not a concern --- I squash the commits before merging them anyway.

src/Polysemy/ConstraintAbsorber.hs
adamConnerSax: Done.  Sorry, I usually check for unnec. pragmas and forgot.  Do we need the plugin bit in each file?

adamConnerSax: Done.  Sorry, I usually check for unnec. pragmas and forgot.  Do we need the plugin bit in each file?

src/Polysemy/ConstraintAbsorber/MonadError.hs
adamConnerSax: It might be.  But I hate it.  I could be convinced.  And it doesn't really matter since we aren't exporting it.  ```Orphan``` speaks to part of it's function but not all of it.  And ```Action``` speaks to a different part.  Both seem better than ```Lift``` to me.  ```OrphanAction```?

src/Polysemy/ConstraintAbsorber/MonadError.hs
adamConnerSax: We ```unsafeCoerce``` it in ```absorbUsingSem```

adamConnerSax: That being said, my sense of how this all works is fairly tenuous.  So I hope it makes sense to you!

src/Polysemy/ConstraintAbsorber/MonadWriter.hs
adamConnerSax: We could, I think.  But I'm worried there will come a time when someone needs to do something else there, like a more complex entailment or something and not having access to that argument will be annoying.  But I don't know since my brain frizzes at ```Sub Dict`` ` in the first place.  Should I try to have two versions of ```absorbWithSem```, one simplified and one where you can access that argument just in case?

adamConnerSax: Actually, right now I can't make that work anyway.  Once I check in the rest of these fixes and you take a look, we can decide if it's worth chasing.

src/Polysemy/ConstraintAbsorber.hs
adamConnerSax: done

src/Polysemy/ConstraintAbsorber.hs
adamConnerSax: Done.  Sorry, I usually check for unnec. pragmas and forgot.  Do we need the plugin bit in each file?

isovector: The plugin is enabled globally across all of the zoo, so no need for it anywhere.

src/Polysemy/ConstraintAbsorber/MonadError.hs
isovector: I'm not going to fight you on this! I just name newtypes `Orphan` whenever they're necessary to avoid orphan instances and I'm not going to export them anyway!

src/Polysemy/ConstraintAbsorber.hs
```diff
@@ -0,0 +1,44 @@
...
+import           Data.Reflection (Reifies, reflect) 
+import           Data.Kind (Type, Constraint)
+
+------------------------------------------------------------------------------
+-- | Given a reifiable constraint @p,
+-- discharge the constraint by providing a dictionary, and a "reified instance"
+-- (often @Sub Dict@ will do)
+-- Specialized to the case when m ~ Sem r
+absorbWithSem :: forall (p :: (Type -> Type) -> Constraint) -- ^ Monadic Constraint to be absorbed
+                        (x :: (Type -> Type) -> Type -> Type ->Type) -- ^ wrapper to avoid orphan instances
```
isovector: Missing a space here before the final `Type`

src/Polysemy/ConstraintAbsorber/MonadError.hs
adamConnerSax: Let's leave it then.  Especially since we can always change it without affecting anything else at all.

src/Polysemy/ConstraintAbsorber.hs
isovector: I'm also not sure what this kind corresponds to. I'd expect it to be `(Type -> Type) -> Type -> Type`. Where does the extra `Type` come from? Is it the `s` in `MonadState s`?

adamConnerSax: So true.  Fixed.

adamConnerSax: It's the ```s``` you need for Reflection.  The phantom type.

adamConnerSax:
Also, while we're changing so much, should these things be
```absorbReader``` or ```absorbMonadReader``` ?  It's clear enough here but not all monadic constraints are named ```MonadXXX``, e.g., ```PandocMonad``` or ```Resource``` or whatever.

isovector:
I have no strong opinions on the naming. I slightly like `absorbReader` better due to the brevity.

src/Polysemy/ConstraintAbsorber/MonadWriter.hs
isovector: Sounds good. I'm not convinced it's possible --- `Dict`s are weird --- but it's a noble goal if tenable.

adamConnerSax: "Dicts are weird" seems like a good first album name for a very very geeky punk band.

isovector:
Does this fix #24 when it lands?

adamConnerSax:
It doesn't fix it because it wasn't broken.  I'll close that one since I found my way through.  If we can think of a couple of more standard (than PandocMonad) examples with superclasses or other non-trivial complications, it might be worth putting them in the zoo just so there are more examples for people to look at.  ```Reflection``` and ```Constraints``` are both confusing so using them together is...a lot.

isovector:
I'm hesitant about `PandocMonad` due to #16 :)

isovector:
I'd like to document the fact that this technology exists in `polysemy`. Do you mind if I write a blog post on it, or do you feel like it would be stealing your thunder? Of course, I'll sing your praise throughout such a post!

adamConnerSax:
Yeah.  I was not suggesting ```PandocMonad``` be allowed anywhere near the zoo!  Just that we keep an eye out for other examples that might provide inspiration.  
```PandocMonad``` was trickier, by a bit, than the reflection example, since the functionality in the PandocMonad part of the dictionary did not cover the MonadError stuff so the dictionary had to be expanded, etc. 
So if we can find a low-dependency-weight example that seems useful in the zoo, it might be worth adding just for educational reasons.

adamConnerSax:
Feel free to blog!  You figured out much of it anyway.  I think I just made you do it...
Someday, I'd like to write about these things as well as you do but that day is not now.
I was hoping you'd write a post about it since it will get people using it and maybe building more absorbers.

adamConnerSax:
Also, on an unrelated note, I never remember what "cargo-cult" means.  So I l[ooked it up](https://en.wikipedia.org/wiki/Cargo_cult) and thought "that makes perfect sense!" and then "and that might be offensive to some since the origin of the term is colonialist condescension to native peoples." and, without much looking, that turns out to be [exactly how some people feel](https://twitter.com/justkelly_ok/status/969641245499801600?lang=en).  So I think I'll not use it--someone suggested "Religion without faith" instead, which works but is less catchy. 
Thought it was worth pointing out since I had no idea!

adamConnerSax:
Not sure if you're waiting for my okay on this, but in case you are, it's good to go from my end.  And I have an updated version of ```polysemy-RandomFu``` ready to go once this is on hackage.

isovector:
Fantastic, thanks!

1atAlcone (approved):

PR #28, opened 2019-06-21
RPC effect

isovector:
Template Haskell for generating RPC interpretations of effects.


isovector:
Looks like we can derive this stuff via kind-generics: https://gitlab.com/trupill/kind-generics/blob/master/kind-generics-deriving/src/Generics/Kind/Derive/Json.hs

PR #35, opened 2019-07-07
Add Final effect

KingoftheHomeless:
This is the `Lift'`/`Final` effect originally discussed at https://github.com/polysemy-research/polysemy/issues/163, but touched up, and with the introduction of a `Tactics`-like environment to make it easier to write interpreters using it.

I'm going to soon add a few more modules to showcase how this can be used, but you should be able to get the gist from `Polysemy.Final`


KingoftheHomeless:
Oh! This hasn't been updated to use the changed names for `Yo`. Perhaps we should resolve that before merging in?

KingoftheHomeless:
Alright, I've now added a slew of uses of `Final`.

Now I'll write some tests and see how everything works.

src/Polysemy/Final/Error.hs
```diff
@@ -0,0 +1,52 @@
...
+  ) where
+
+import Control.Exception
+import Data.Typeable
+import Polysemy
+import Polysemy.Final
+import Polysemy.Error hiding (throw, catch)
+
+------------------------------------------------------------------------------
+-- | Run an 'Error' effect as an 'IO' 'Exception'.
```
isovector: Same warning about semantics, please

src/Polysemy/Final/Error.hs
```diff
@@ -0,0 +1,52 @@
...
+  Throw e   -> pure $ throwIO $ WrappedExc e
+  Catch m h -> do
+    m' <- runS m
+    h' <- bindS h
+    s  <- getInitialStateS
+    pure $ catch m' $ \(se :: WrappedExc e) ->
+      h' (unwrapExc se <$ s)
+
+
+newtype WrappedExc e = WrappedExc { unwrapExc :: e }
```
isovector: We can just expose this stuff in polysemy if it's easier for you!

src/Polysemy/Final/Fixpoint.hs
```diff
@@ -0,0 +1,20 @@
...
+
+import Polysemy
+import Polysemy.Final
+import Polysemy.Fixpoint
+
+import Control.Monad.Fix
+
+-----------------------------------------------------------------------------
+-- | Run a 'Fixpoint' effect through a final 'MonadFix'
+runFixpointFinal :: (Member (Final m) r, MonadFix m)
```
isovector: This one I like. I suspect it will diverge less than the existing fixpoint stuff :)

src/Polysemy/Final/MTL.hs
```diff
@@ -0,0 +1,73 @@
...
+
+import Polysemy
+import Polysemy.Final
+import Polysemy.Error hiding (throw, catch)
+import Polysemy.Reader hiding (ask, local)
+import Polysemy.State hiding (get, put)
+import Polysemy.Writer hiding (tell, listen, censor)
+
+-----------------------------------------------------------------------------
+-- | Run an 'Error' effect through a final 'MonadError'
```
isovector: Here and elsewhere, please add the same warnings about semantics.

src/Polysemy/Final.hs
```diff
@@ -0,0 +1,279 @@
...
+-- but be warned, this breaks the implicit contract of @'LastMember' ('Lift' m)@
+-- that @m@ truly _is_ the final monad, so depending on the final monad and operations
+-- used, 'runFinalTrans' may become _unsafe_.
+--
+-- For example, 'runFinalTrans' is unsafe with 'runAsync' if
+-- the final monad is non-deterministic, or a continuation
+-- monad.
+runFinalLift :: Monad m
+              => (forall x. n x -> m x)
+              -> Sem [Final m, Lift m, Lift n] a
```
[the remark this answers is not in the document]
KingoftheHomeless: Well, the reasoning was that if you had some `Final` monad that isn't `IO`, but is `MonadIO`, you still couldn't use ops with a `LastMember (Lift IO)` constraint.
In order to use stuff like `runAsync`, the last member of the effect stack must be `Lift IO`, but you can't normally have that if the `Final` monad isn't `IO`.
This is for a very good reason, as `runAsync` typically isn't safe when the end monad isn't directly `IO`, but in some cases, it might still work out fine. `runFinalLift` is for those cases.

However, the only convincing motivational example I had at the time now turns out to be unsafe, so the only things I can point to are simple monad transformers like having `Final (WriterT IO)` but `LastMember (Lift IO)`, in which case using `runFinalLiftIO` together with `runWriterFinal` and `runAsync` _is_ safe... probably.

But that's not very convincing of a use-case. I am willing to drop `runFinalLift` and `runFinalLiftIO`.

The extra `Lift` for `m` is for the sake of consistency, and the ability for interpreters to use `sendM` and not just `withWeaving` and friends on the target monad.

src/Polysemy/Final/Error.hs
KingoftheHomeless: That would be nice. Put it in an internal module.

src/Polysemy/Final/Fixpoint.hs
KingoftheHomeless: State semantics aren't relevant when it comes to `mfix` right? So I can elide the **Beware**.

src/Polysemy/Final.hs
```diff
@@ -0,0 +1,279 @@
...
+               )
+            -> Sem r a
+
+-----------------------------------------------------------------------------
+-- | 'withWeaving' admits an implementation of 'sendM'
+-- by providing the means of threading effects through 'Sem r x'
+-- to the final monad.
+--
+-- Just like 'sendM', you are discouraged from using this in application code.
+sendM' :: Monad m => Member (Final m) r => m a -> Sem r a
```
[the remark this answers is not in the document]
KingoftheHomeless: Yeah, I just stuck with `Monad` for consistency's sake. I'll relax constraints where possible.

KingoftheHomeless:
From what I can tell, "-Final" interpreters act exactly like the kind of interpreters that take explicit lowering actions, such as `runAsyncInIO` and `runErrorInIO`, only more flexible, as they work through weavings, which can be provided by `Sem` implicitly, rather than straight-up natural transformations.

The difference is that you can place them `-Final` interpreters anywhere, whereas interpreters with explicit lowering functions need to run after return-value changing effects in order to be composed using `.@`.

You _could_ actually place such interpreters after return-value changing effects by using `fmap` and the like to massage the interpreter chain into a natural transformation, in which case these interpreters will exhibit the same strange semantics as  "-Final" interpreters (as effects that care about state semantics typically have return-value changing interpreters.)

"Final", in a sense, makes it a bit "too easy" to do that kind of evil interpreter chaining, but has a much better UX. Providing lowering functions is very verbose, and working with '.@' is rather finicky. Because of that, I'd like for "Final" to eventually become part of Polysemy proper, but at the moment, it can live here.

src/Polysemy/Final.hs
```diff
@@ -0,0 +1,279 @@
...
+  in
+    go
+{-# INLINE interpretHFinal #-}
+
+------------------------------------------------------------------------------
+-- | 'Strategic' is an environment in which you're capable of explicitly
+-- threading higher-order effect states to the final monad.
+-- This is based upon @Tactics@, (see 'Tactical'), and usage
+-- is extremely similar.
+type Strategic m n a = forall f. Functor f => Sem (WithStrategy m f n) (m (f a))
```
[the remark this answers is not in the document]
KingoftheHomeless: Thanks!

Actually, maybe its _too_ solid of a pun. Would you like to reserve the name for the future?

KingoftheHomeless:
By the way, the tests in `FinalIOSpec` work through `IO` in order to test how `-Final` interpreters work within it. Is that OK with you?

test/FinalIOSpec.hs
```diff
@@ -0,0 +1,92 @@
...
+follow (Node _ ref) = sendM $ readIORef ref
+
+test1 :: IO (Either Int (String, Int, Maybe Int))
+test1 = do
+  ref <- newIORef "abra"
+  runFinal
+    . runStateInIORef ref -- Order of these interpreters don't matter
+    . runErrorInIOFinal
+    . runFixpointFinal
+    . runAsyncFinal
```
KingoftheHomeless: This is an example of a scenario where you can't use `runAsync`. For some reason, the combination of exceptions and fixpoint results in it being blocked indefinitely on a `MVar`. (It doesn't matter if you replace the other `-Final` interpreters with their non-`Final `counterparts.)

src/Polysemy/Final/Fixpoint.hs
isovector: I think so!

KingoftheHomeless: Then that's what I'll do.

isovector:
Let me know when this is ready for me to take a look!

KingoftheHomeless:
It's ready now; I've added tests. Although, I do need to update this to use the new name for `Yo`. I'd wait until #36 is resolved before merging this in, as in resolving that issue we will work out how to properly update polysemy-zoo.

KingoftheHomeless:
Perhaps you want me to ping you? @isovector

KingoftheHomeless:
"It's ready", I say, as I notice "hey we can do stuff with resource too".

KingoftheHomeless:
Rebasing is a nefarious beast. Anyway, what's your opinion on merging this, @isovector ? 
We still have the unresolved question on whether or not we should remove `runFinalLift` and `runFinalLiftIO`.

isovector:
I'll merge anything that's green into the zoo! Here be dragons!

KingoftheHomeless:
Then let's do it.

Issue #36, opened 2019-07-08
Update absorbWriter to the new Writer effect


isovector:
https://github.com/polysemy-research/polysemy/pull/169

isovector:
@adamConnerSax do you mind picking up this ticket?

KingoftheHomeless:
This can be closed now.

PR #37, opened 2019-07-08
Updated AbsorbMonadWriter

adamConnerSax:
- updated package.yaml and cabal for polysemy >= v0.7.0.0
- re-implemented ```semPass``` in terms of ```Polysemy.Writer.pass``` instead of ```Polysemy.Writer.censor```
- confirmed all tests pass.  NB: one failed after this change when the implementation of ```semPass``` was still via censor.  So the tests are sensitive to the change of semantics.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -31,3 +31,5 @@
 
 ## Unreleased changes
 
+* In AbsorbMonadWriter, re-implemented mtl pass in terms of the new Polysemy.Writer
+pass and confirmed all tests.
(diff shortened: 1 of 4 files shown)


adamConnerSax:
Resolves https://github.com/polysemy-research/polysemy-zoo/issues/36

KingoftheHomeless:
We need to merge this ASAP. polysemy-zoo currently fails to build when you have polysemy-7.0.0.0 due to the package.yaml having incorrect dependencies.

adamConnerSax:
Huh.  That's odd.  I didn't have to update anything (except the polysemy bound itself) when I did this PR.  Is it just the polysemy bound itself that fails?

KingoftheHomeless:
I noticed it from Travis, but maybe Travis is just incorrectly configured?

KingoftheHomeless:
That must be it, because I have to coerce stack to use 0.7.0.0. So I'm wrong, there's no hurry. Sorry!

isovector:
Excellent, thanks!

PR #39, opened 2019-07-10
Fix docs and changelog for Final, add inlining to Final functions

KingoftheHomeless:


src/Polysemy/Final.hs
```diff
@@ -267,10 +271,12 @@ runFinalLift nat = usingSem $ \u -> case decomp u of
     Right (Weaving (Lift m) s _ ex _) -> ex . (<$ s) <$> m
     Left g' -> case extract g' of
       Weaving (Lift n) s _ ex _ -> ex . (<$ s) <$> nat n
+{-# INLINE runFinalLift #-}
 
 ------------------------------------------------------------------------------
--- | 'runFinalTrans', specialized to transform 'IO' to a 'MonadIO'.
+-- | 'runFinalLift', specialized to transform 'IO' to a 'MonadIO'.
```
isovector: `runFinalLiftIO`

polysemy-zoo.cabal
```diff
@@ -99,7 +99,7 @@ test-suite polysemy-zoo-test
     , hedis >=0.10 && <0.13
     , hspec >=2.6.0 && <3
     , mtl >=2.0.1.0 && <3.0.0.0
-    , polysemy >=0.5.1.0
+    , polysemy >=0.7.0.0
```
isovector: Nice catch! I don't know how I missed this, but it probably breaks stack. Whoops!

src/Polysemy/Final.hs
isovector: Nevermind. I understand what's happening here. Maybe I shouldn't do reviews at 11pm!

PR #42, opened 2019-07-24
Documentation fixes for Continuation effects

KingoftheHomeless:
While digging through the files I noticed a rather embarrassing mistake in the documentation. Given how prone I am to those kinds of mistakes, I suppose documentation isn't my strong suit.


isovector:
Nice catch, thanks!

PR #43, opened 2019-07-24
Update resolver and bounds, fix uses of makeSem_

KingoftheHomeless:
Turns out @Avi-D-coder by mistake didn't bring in my fixes of `makeSem_`, so I'm creating this pull request to bring them in.

I've also bumped up the resolver to a version I'm sure this will actually work on.


Avi-D-coder:
Oops, I must have checked out the wrong commit from that branch.

KingoftheHomeless:
I noticed that the bounds for polysemy was also incorrect, which may also be the reason that the documentation is borked on Hackage.

isovector:
Excellent, thanks!

PR #45, opened 2019-08-17
Final rework, Fresh Effect, Cont ConstraintAbsorber and new Cont interpreter

KingoftheHomeless:
This is a big one.

@isovector You'll probably find `runContViaFresh` (previously `runContViaCapture`) even more interesting now, because I found a promising way to solve the problems it had that I described to you! The problems `Capture` has still probably remains impossible to solve, though.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -36,7 +36,6 @@ dependencies:
 - async >= 2.2 && < 3
 - binary >= 0.8.5.1 && < 0.9
 - bytestring >= 0.10.8.2 && < 0.11
-- hashable < 2
 
 
 ghc-options:
--- a/package.yaml
+++ b/package.yaml
@@ -24,10 +24,12 @@ dependencies:
 - constraints >= 0.10.1 && < 0.12
 - containers >= 0.5 && < 0.7
 - mtl >= 2.0.1.0 && < 3.0.0.0
-- polysemy >= 1.0.0.0
+- stm >= 2 && < 3
+- polysemy >= 1.1.0.0
 - polysemy-plugin >= 0.2
 - random >= 1.1 && <1.2
 - reflection >= 2.1.4 && < 3.0.0
+- transformers >= 0.5.2.0 && < 0.6
 - ghc-prim >= 0.5.2 && < 0.6
 - hedis >= 0.10 && < 0.13
 - async >= 2.2 && < 3
@@ -72,7 +74,7 @@ tests:
     build-tools:
     - hspec-discover  >= 2.0
     dependencies:
-    - polysemy >= 0.7.0.0
+    - polysemy >= 1.1.0.0
     - polysemy-zoo
     - hspec >= 2.6.0 && < 3
 
--- a/package.yaml
+++ b/package.yaml
@@ -36,6 +36,7 @@ dependencies:
 - async >= 2.2 && < 3
 - binary >= 0.8.5.1 && < 0.9
 - bytestring >= 0.10.8.2 && < 0.11
+- hashable < 2
 
 
 ghc-options:
--- a/package.yaml
+++ b/package.yaml
@@ -23,6 +23,7 @@ dependencies:
 - base >= 4.9 && < 5
 - constraints >= 0.10.1 && < 0.12
 - containers >= 0.5 && < 0.7
+- contravariant < 2
 - mtl >= 2.0.1.0 && < 3.0.0.0
 - stm >= 2 && < 3
 - polysemy >= 1.1.0.0
--- a/src/Polysemy/Cont.hs
+++ b/src/Polysemy/Cont.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE Trustworthy #-}
 module Polysemy.Cont
   (-- * Effect
     Cont(..)
@@ -10,7 +11,10 @@ module Polysemy.Cont
     -- * Interpretations
   , runContPure
   , runContM
-  , runContFinal
+  , contToFinal
+
+    -- * Experimental Interpretations
+  , runContViaFresh
 
     -- * Unsafe Interpretations
   , runContUnsafe
@@ -18,6 +22,7 @@ module Polysemy.Cont
   -- * Prompt types
   , Ref(..)
   , ExitRef(..)
+  , ViaFreshRef
   ) where
 
 import Data.Void
@@ -27,7 +32,10 @@ import Polysemy.Final
 
 import Polysemy.Cont.Internal
 
-import Control.Monad.Cont (MonadCont())
+import Polysemy.Error
+import Polysemy.Fresh
+
+import Control.Monad.Cont (MonadCont(), ContT(..), runContT)
 import qualified Control.Monad.Cont as C (callCC)
 
 -----------------------------------------------------------------------------
@@ -35,11 +43,8 @@ import qualified Control.Monad.Cont as C (callCC)
 -- Executing the provided continuation will abort execution.
 --
 -- Using the provided continuation
--- will rollback all effectful state back to the point where 'callCC' was invoked,
--- unless such state is interpreted in terms of the final
--- monad, /or/ the associated interpreter of the effectful state
--- is run after 'runContUnsafe', which may be done if the effect isn't
--- higher-order.
+-- will rollback all local effectful state back to the point where
+-- 'callCC' was invoked.
 --
 -- Higher-order effects do not interact with the continuation in any meaningful
 -- way; i.e. 'Polysemy.Reader.local' or 'Polysemy.Writer.censor' does not affect
@@ -47,8 +52,8 @@ import qualified Control.Monad.Cont as C (callCC)
 -- The only exception to this is if you interpret such effects /and/ 'Cont'
 -- in terms of the final monad, and the final monad can perform such interactions
 -- in a meaningful manner.
-callCC :: forall ref a r.
-          Member (Cont ref) r
+callCC :: forall ref r a
+       .  Member (Cont ref) r
        => ((forall b. a -> Sem r b) -> Sem r a)
        -> Sem r a
 callCC cc = subst (\ref -> cc (jump ref)) pure
@@ -77,11 +82,11 @@ runContM = runContUnsafe
 --
 -- /Beware/: Effects that aren't interpreted in terms of the final monad
 -- will have local state semantics in regards to 'Cont' effects
--- interpreted this way. See 'interpretFinal'.
-runContFinal :: (Member (Final m) r, MonadCont m)
-             => Sem (Cont (ExitRef m) ': r) a
-             -> Sem r a
-runContFinal = interpretFinal $ \case
+-- interpreted this way. See 'Final'.
+contToFinal :: (Member (Final m) r, MonadCont m)
+            => Sem (Cont (ExitRef m) ': r) a
+            -> Sem r a
+contToFinal = interpretFinal $ \case
   Jump ref a    -> pure $ enterExit ref a
   Subst main cb -> do
     main' <- bindS main
@@ -89,7 +94,40 @@ runContFinal = interpretFinal $ \case
     s     <- getInitialStateS
     pure $ C.callCC $ \exit ->
       main' (ExitRef (\a -> cb' (a <$ s) >>= vacuous . exit) <$ s)
-{-# INLINE runContFinal #-}
+{-# INLINE contToFinal #-}
+
+-----------------------------------------------------------------------------
+-- | A highly experimental 'Cont' interpreter that functions
+-- through a combination of 'Error' and 'Fresh'. This may be used safely
+-- anywhere in the effect stack.
+--
+-- 'runContViaFresh' is still under development.
+-- You're encouraged to experiment with it, but don't rely on it.
+-- For best results, use 'runContViaFresh' as the first interpreter you run,
+-- such that all other effects are global in respect to it.
+--
+-- This interpreter may return 'Nothing' if the control flow becomes
+-- split into separate, inconsistent parts,
+-- such that backtracking fails when trying to invoke continuations.
+-- For example, if you reify a continuation inside an
+-- 'async':ed thread, and then have that thread return the reified
+-- continuation back to the main thread through an 'await', then
+-- 'runContViaFresh' will return 'Nothing' upon executing the continuation
+-- in the main thread.
+runContViaFresh :: forall uniq r a
+                 . (Member (Fresh uniq) r, Eq uniq)
+                => Sem (Cont (ViaFreshRef uniq) ': r) a
+                -> Sem r (Maybe a)
+runContViaFresh =
+  let
+    hush (Right a) = Just a
+    hush _         = Nothing
+  in
+      fmap hush
+    . runError
+    . (`runContT` pure)
+    . runContViaFreshInC
+{-# INLINE runContViaFresh #-}
 
 -----------------------------------------------------------------------------
 -- | Runs a 'Cont' effect by providing 'pure' as the final continuation.
@@ -99,8 +137,8 @@ runContFinal = interpretFinal $ \case
 -- 'Polysemy.Writer.censor' will be no-ops, 'Polysemy.Error.catch' will fail
 -- to catch exceptions, and 'Polysemy.Writer.listen' will always return 'mempty'.
 --
--- __You should therefore use 'runContUnsafe' /after/ running all interpreters for
--- your higher-order effects.__
+-- __You should therefore use 'runContUnsafe' only /after/ running all__
+-- __interpreters for your higher-order effects.__
 --
 -- Note that 'Final' is a higher-order effect, and thus 'runContUnsafe' can't
 -- safely be used together with 'runFinal'.
--- a/src/Polysemy/Cont/Internal.hs
+++ b/src/Polysemy/Cont/Internal.hs
@@ -147,8 +147,7 @@ instance Contravariant (ViaFreshRef uniq) where
   continuation /it/ gets, but can't provide to the higher-order computation.
 
   I'm astonished that this even remotely works, but it does have some rather
-  weird behaviour I haven't completely figured out yet; especially in the
-  interaction with effects local in regards to the 'Cont'.
+  weird behaviour I haven't completely figured out yet.
 
   I'm reasonably happy with how 'runContViaFreshInC' looks;
   I'm a lot less happy with 'runContViaFreshInCWeave', I just kinda threw
--- a/src/Polysemy/Cont/Internal.hs
+++ b/src/Polysemy/Cont/Internal.hs
@@ -1,11 +1,19 @@
-{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TemplateHaskell, Unsafe #-}
 module Polysemy.Cont.Internal where
+
+import Data.Functor.Contravariant
 
 import Polysemy
 import Polysemy.Internal
 import Polysemy.Internal.Union
+import Polysemy.Fresh
+import Polysemy.Error
+
 import Control.Monad
-import Control.Monad.Cont (ContT(..))
+import Control.Monad.Trans.Cont hiding (Cont)
+
+import Unsafe.Coerce
+import GHC.Exts (Any)
 
 -----------------------------------------------------------------------------
 -- | An effect for abortive continuations.
@@ -63,7 +71,7 @@ subst :: forall ref a b r
 -- 'Polysemy.Writer.censor' will be no-ops, 'Polysemy.Error.catch' will fail
 -- to catch exceptions, and 'Polysemy.Writer.listen' will always return 'mempty'.
 --
--- __You should therefore use 'runContUnsafeWithC' /after/ running all interpreters
+-- __You should therefore use 'runContWithCUnsafe' /after/ running all interpreters
 -- for your higher-order effects.__
 runContWithCUnsafe :: (a -> Sem r s) -> Sem (Cont (Ref (Sem r) s) ': r) a -> Sem r s
 runContWithCUnsafe c (Sem m) = (`runContT` c) $ m $ \u -> case decomp u of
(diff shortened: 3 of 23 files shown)


src/Polysemy/Final.hs
```diff
@@ -208,77 +169,86 @@ liftS m = do
...
--- | Lower a 'Sem' containing only a lifted, final monad into that monad.
--- The appearance of 'Lift' as the final effect
--- is to allow the use of operations that rely on a @'LastMember' ('Lift' m)@
--- constraint.
-runFinal :: Monad m => Sem '[Final m, Embed m] a -> m a
-runFinal = usingSem $ \u -> case decomp u of
-  Right (Weaving (WithWeaving wav) s wv ex ins) ->
+-- | Lower a 'Sem' containing only a single lifted, final 'Monad' into that
+-- monad.
+runFinal :: Monad m => Sem '[Final m] a -> m a
```
KingoftheHomeless: I removed the `Embed` here, since `LastMember` doesn't exist anymore, and you can use `embedToFinal` instead of processing both effects simultaneously.

That said, given that most uses of `runFinal` will be paired with a `embedToFinal` because of this, perhaps `runFinal` should just eliminate an `Embed` effect too while it's at it. It's probably more efficient that way too. Your thoughts?

KingoftheHomeless:
Everything `Final` here is what I'm planning should move to `Polysemy`, except `Final` with `Cont`, the `-FinalGlobal` interpreters (too experimental), and `Polysemy.Final.MTL` (I feel `MTL` stuff can stay in `polysemy-zoo`).

isovector:
Thanks for the PR! Might take me a few days to work through it though :)

src/Polysemy/Fresh.hs
```diff
@@ -0,0 +1,115 @@
...
+-----------------------------------------------------------------------------
+-- | An effect for creating unique objects which may be used as references,
+-- a la 'Unique'. Polymorphic code making use of 'Fresh' is expected
+-- to place constraints upon @uniq@ as necessary.
+--
+-- Any interpreter for 'Fresh' has the responsibilty of ensuring
+-- that any call to 'fresh' produces an object that /never/
+-- compares equal to an object produced by a previous call to 'fresh'.
+data Fresh uniq m a where
+  Fresh :: Fresh uniq m uniq
```
KingoftheHomeless: I'm deviating from `fused-effects` here in that I don't provide a `reset` action. My intention with this effect is that it is a unique symbol generator; nothing more, and definitely nothing less. Having `reset` would contradict that, and is also the reason why the pure interpreter (`runFreshUnsafe`) is marked as unsafe, since the state could get backtracked by another effect/the final monad, so `fresh` in those cases doesn't guarantee that the symbol it generates is unique.

src/Polysemy/Final.hs
KingoftheHomeless: Personally I prefer having `runFinal` only interpret `Final`.

src/Polysemy/Cont/Internal.hs
```diff
@@ -95,4 +103,155 @@ embedSem = liftSem . weave (pure ()) (pure . join) inspectSem
...
+  we instead use 'runContViaFreshInCWeave'; this makes use 'ContFreshState'
+  as its functorial state, which stores /handlers/ for backtrack exceptions.
+  'runContViaFreshInCWeave', in addition to 'catch'ing exceptions on the continuation
+  it is given, /also/ returns the handler it uses for the 'catch'.
+  This handler is then used by 'runContViaFresh' to catch exceptions on the
+  continuation /it/ gets, but can't provide to the higher-order computation.
+
+  I'm astonished that this even remotely works, but it does have some rather
+  weird behaviour I haven't completely figured out yet; especially in the
+  interaction with effects local in regards to the 'Cont'.
```
KingoftheHomeless: This turns out to have been due to the issue https://github.com/polysemy-research/polysemy/pull/210 addresses. Still, the exact semantics of `runContViaFresh` still hasn't been figured out yet.

src/Polysemy/Cont.hs
```diff
@@ -77,19 +82,52 @@ runContM = runContUnsafe
...
+--
+-- 'runContViaFresh' is still under development.
+-- You're encouraged to experiment with it, but don't rely on it.
+-- For best results, use 'runContViaFresh' as the first interpreter you run,
+-- such that all other effects are global in respect to it.
+--
+-- This interpreter may return 'Nothing' if the control flow becomes
+-- split into separate, inconsistent parts,
+-- such that backtracking fails when trying to invoke continuations.
+-- For example, if you reify a continuation inside an
```
isovector: Dang, that'd be super cool if it worked!

src/Polysemy/Cont/Internal.hs
```diff
@@ -95,4 +103,155 @@ embedSem = liftSem . weave (pure ()) (pure . join) inspectSem
...
+  then nothing can catch the underlying backtrack exception
+  once it is thrown.
+
+  The solution to this is anothor kludge: when weaving other effects through,
+  we instead use 'runContViaFreshInCWeave'; this makes use 'ContFreshState'
+  as its functorial state, which stores /handlers/ for backtrack exceptions.
+  'runContViaFreshInCWeave', in addition to 'catch'ing exceptions on the continuation
+  it is given, /also/ returns the handler it uses for the 'catch'.
+  This handler is then used by 'runContViaFresh' to catch exceptions on the
+  continuation /it/ gets, but can't provide to the higher-order computation.
```
isovector: This is a very clever idea. Thanks for the writeup!

src/Polysemy/Cont/Internal.hs
```diff
@@ -95,4 +103,155 @@ embedSem = liftSem . weave (pure ()) (pure . join) inspectSem
...
+
+instance Contravariant (Ref m s) where
+  contramap f ref = Ref (runRef ref . f)
+
 newtype ExitRef m a = ExitRef { enterExit :: forall b. a -> m b }
+
+instance Contravariant (ExitRef m) where
+  contramap f ref = ExitRef $ \a -> enterExit ref (f a)
+
+data ViaFreshRef uniq a = ViaFreshRef { getBacktrackException :: a -> (uniq, Any) }
```
isovector: What is this `Any`? Could we existentialize whatever it is instead of unsafecoercing?

src/Polysemy/Cont/Internal.hs
```diff
@@ -95,4 +103,155 @@ embedSem = liftSem . weave (pure ()) (pure . join) inspectSem
...
+            loop act =
+              runContT (ex <$> act) c `catch` \ x@(ref', a') -> do
+                if ref == ref' then
+                  loop (cn' $ unsafeCoerce a')
+                else
+                  throw x
+          ResAndHandler res h <-
+            loop $ main' $ ViaFreshRef (\a -> (ref, unsafeCoerce a))
+          return $ ResAndHandler res
+              -- TODO(KingoftheHomeless): This handler is dubious.
```
isovector: How so?

src/Polysemy/Final.hs
```diff
@@ -36,107 +36,77 @@ module Polysemy.Final
...
 -- interpret higher-order effects in terms of the final monad.
 --
--- /Beware/: Any interpreters built using this (or 'Final' in general)
--- will /not/ respect local/global state semantics based on the order of
--- interpreters run. You should signal interpreters that make use of
--- 'Final' by adding a "-Final" suffix to the names of these.
+-- 'interpretFinal' requires less boilerplate than using 'interpretH'
+-- together with 'withStrategic'\/'withWeaving', but is also less powerful.
+-- 'interpretFinal' does not provide any means of executing actions
+-- of @'Sem' r@ as you interpret each action, and the provided interpreter
```
isovector: Tangentially, @ChrisPenner and I were looking for an action `Member (Final m) r => (Sem r a -> m a) -> Sem r a` the other day. Can we do that with `Final`? This comment suggests yes, but I couldn't find any actual code that did it.

src/Polysemy/Final.hs
isovector: I agree

src/Polysemy/Final.hs
```diff
@@ -208,77 +169,86 @@ liftS m = do
...
--- | Lower a 'Sem' containing two lifted monad into the final monad,
--- by interpreting one of the monads in terms of the other one.
---
--- This allows for the use of operations that rely on a @'LastMember' ('Lift' m)@
--- constraint, as long as @m@ can be transformed to the final monad;
--- but be warned, this breaks the implicit contract of @'LastMember' ('Lift' m)@
--- that @m@ /is/ the final monad, so depending on the final monad and operations
--- used, 'runFinalLift' may become /unsafe/.
+-- | Run a 'Sem' containing only @'Final' 'Identity'@ as a pure value.
+runFinalPure :: forall a. Sem '[Final Identity] a -> a
```
isovector: Can we somehow unify this with `run`?

src/Polysemy/Final.hs
```diff
@@ -208,77 +169,86 @@ liftS m = do
...
+runFinalPure = coerce (runFinal @Identity @a)
+{-# INLINE runFinalPure #-}
+
+------------------------------------------------------------------------------
+-- | Given natural transformations between @m1@ and @m2@, run a @'Final' m1@
+-- effect by transforming it into a @'Final' m2@ effect.
+finalToFinal :: forall m1 m2 a r
```
isovector: What's the use case for this? The isomorphism between `m1` and `m2` restricts any case I can think of.

src/Polysemy/Final.hs
```diff
@@ -208,77 +169,86 @@ liftS m = do
...
-    Left g' -> case extract g' of
-      Weaving (Embed n) s _ ex _ -> ex . (<$ s) <$> nat n
-{-# INLINE runFinalLift #-}
+    ex <$> wav s (runFinalSem . wv) ins
+  Left g -> liftSem (hoist runFinalSem g)
+{-# INLINE runFinalSem #-}
+
+------------------------------------------------------------------------------
+-- | Run a @'Final' m@ effect by providing an explicit lowering function.
+lowerFinal :: Member (Embed m) r
```
isovector: How many times does `f` get called here? Is it one for each action in `Final`, or is it one time total?

src/Polysemy/Final/Async.hs
```diff
@@ -18,21 +20,42 @@ import Polysemy.Final
...
 -- State that seems like it should be threaded globally throughout the `Async`
 -- /will not be./
 --
--- Prefer 'asyncToIO' unless its unsafe or inefficient in the context of your
--- application.
-runAsyncFinal :: Member (Final IO) r
-              => Sem (Async ': r) a
-              -> Sem r a
-runAsyncFinal = interpretFinal $ \case
+-- Prefer 'asyncToIO' or 'asyncToIOThreadless' unless these are unsafe or
```
isovector: I know you have the comment above about `lowerAsync`, but I think it warrants suggesting to prefer this over `lowerAsync`; my understanding is we can deprecate that thing when this lands?

src/Polysemy/Final/Async.hs
```diff
@@ -18,21 +20,42 @@ import Polysemy.Final
...
+    liftS $ A.async (inspect ins <$> m')
+  Await a -> liftS (A.wait a)
+{-# INLINE asyncToIOFinal #-}
+
+------------------------------------------------------------------------------
+-- | 'asyncToIOFinal' implemented using 'interpretFinalGlobal'.
+--
+-- This behaves semantically very much like 'asyncToIO',
+-- but doesn't need to spin up an interpreter thread, making it more
+-- efficient (but not any more safe).
```
isovector: Does no more safety here mean "just as likely to deadlock"?

src/Polysemy/Final/Error.hs
```diff
@@ -13,28 +13,28 @@ import           Polysemy.Final
 import           Polysemy.Error
 
 ------------------------------------------------------------------------------
--- | Run an 'Error' effect as an 'IO' 'Exception'.
+-- | Run an 'Error' effect as an 'IO' 'Exception' through 'Final' 'IO'.
```
isovector: When this PR moves into polysemy proper, is the plan to merge eg `Polysemy.Final.Error` into `Polysemy.Error`?

src/Polysemy/Final/Internal.hs
```diff
@@ -0,0 +1,211 @@
...
+--     together with the initial state @f ()@, and an @'Inspector' f@.
+--     This is powerful enough for most purposes.
+--
+-- /Beware/: 'Final' actions are interpreted as actions of the final monad,
+-- and the effectful state visible to
+-- 'withWeaving'\/'Polysemy.Final.withStrategic'\/'Polysemy.Final.interpretFinal'
+-- is that of /all interpreters run in order to produce the final monad/.
+--
+-- This means that any interpreter built using 'Final' will /not/
+-- respect local/global state semantics based on the order of
```
isovector: To concretize this for my understanding, this is essentially saying "`IORef`s are always global, so you shouldn't expect them to be local," right?

isovector: oh nevermind, you explain immediately after!

src/Polysemy/Final/Internal.hs
```diff
@@ -0,0 +1,211 @@
...
+--
+-- State semantics of effects that are /not/
+-- interpreted in terms of the final monad will always
+-- appear local to effects that are interpreted in terms of the final monad.
+--
+-- State semantics between effects that are interpreted in terms of the final monad
+-- depend on the final monad. For example, if the final monad is a monad transformer
+-- stack, then state semantics will depend on the order monad transformers are stacked.
+newtype Final m z a where
+  WithWeaving :: (  forall f
```
isovector: How about wrapping this scary thing in a GADT so we can name all of its parameters?

src/Polysemy/Final/Internal.hs
```diff
@@ -0,0 +1,211 @@
...
+-- adressing the primary issue with 'Final'.
+--
+-- Semantically, interpreters written using this behave very much as
+-- though they were written using 'withLowerToIO'.
+-- However, this does not need to spawn an interpreter thread, making
+-- it more efficient (but not any more safe.)
+--
+-- 'interpretFinalGlobal' operates under the assumption that any effectful
+-- state which can't be inspected using 'Polysemy.Inspector' can't contain any
+-- values. This is true for all interpreters featured in polysemy,
```
isovector: I'n not sure what it means to not contain any values. The functorial state for `State` contains values, doesn't it?

src/Polysemy/Final/Async.hs
KingoftheHomeless: Basically, yeah. It inherits `https://github.com/polysemy-research/polysemy/issues/205`. I also make no guarantee that this isn't more _unsafe_. There could manifest different issues regarding `interpretFinalGlobal` which doesn't with `withLowerToIO`.

src/Polysemy/Final/Error.hs
KingoftheHomeless: That's my thought, yeah.

src/Polysemy/Final/MTL.hs
```diff
@@ -24,64 +24,64 @@ import Polysemy.Writer hiding (tell, listen, pass)
...
--- /Beware/: Effects that aren't interpreted in terms of the final
+-- /Beware/: Effects that aren't interpreted in terms of the embedded
 -- monad will have local state semantics in regards to 'State' effects
--- interpreted this way. See 'interpretFinal'.
-runStateFinal :: (Member (Embed m) r, MonadState s m)
-               => Sem (State s ': r) a
-               -> Sem r a
-runStateFinal = interpret $ \case
+-- interpreted this way. See 'Final'.
+stateToEmbed :: (Member (Embed m) r, MonadState s m)
```
isovector: How well does this type get inferred? Things like this would always throw `freer-simple` for a loop, but maybe the plugin helps in this case?

src/Polysemy/Fresh.hs
isovector: I'm cool with this. That's all I ever use it for too.

src/Polysemy/Final/MTL.hs
KingoftheHomeless: The plugin is awesome enough to resolve this.

src/Polysemy/Final/Internal.hs
KingoftheHomeless: Yeah, but the functorial state for `State` is always inspectable. Compare this to the functorial state for `Error`: `Either e a`. If it's uninspectable, then it must be a `Left` value, and thus it doesn't contain any values of `a`.

src/Polysemy/Fresh.hs
```diff
@@ -0,0 +1,115 @@
...
+-- sharing the call to 'unsafePerformIO'.
+-- The inlining of 'interpret' is so that I can give
+-- 'u' to 'newUnique'', and thus prevent 'uniq' from floating outside the
+-- lambda. This interpreter might even be safe to INLINE, but I'm not taking
+-- any chances.
+{-# NOINLINE runFreshUnsafePerformIO #-}
+
+newUnique' :: Union (Fresh Unique ': r) (Sem (Fresh Unique ': r)) a -> IO Unique
+newUnique' (Union _ _) = newUnique
+{-# NOINLINE newUnique' #-}
```
isovector: I'd love to see a `runFreshNum :: Num n => Sem (Fresh n ': r) a -> Sem r a` interpretation here

package.yaml
```diff
@@ -23,11 +23,14 @@ dependencies:
...
 - containers >= 0.5 && < 0.7
+- contravariant < 2
 - mtl >= 2.0.1.0 && < 3.0.0.0
-- polysemy >= 1.0.0.0
+- stm >= 2 && < 3
+- polysemy >= 1.1.0.0
 - polysemy-plugin >= 0.2
 - random >= 1.1 && <1.2
 - reflection >= 2.1.4 && < 3.0.0
+- transformers >= 0.5.2.0 && < 0.6
```
isovector: Do we need this?

src/Polysemy/Final/Internal.hs
isovector: I understand, thanks! Mind adding that description to the comment?

src/Polysemy/Fresh.hs
KingoftheHomeless: Are you in favour of generalizing `runFreshUnsafe` to have that signature, or should we have a seperate `runFreshNumUnsafe`? The issue with generalizing `runFreshUnsafe` is that you'd run into ambiguity issues if you don't care about the specific `Num`.

src/Polysemy/Final/MTL.hs
isovector: Even at the call site?

src/Polysemy/Final/Internal.hs
KingoftheHomeless: You mean so it'd become something like this?:
```haskell
newtype Final m z a where
  WithWeaving :: (forall f. Functor f => WithWeavingState f m -> m (f a)) 
                        -> WithWeaving m a
```
Where `WithWeavingState` is a record?

If the goal is to improve documentation, I'd instead prefer this, if it works with Haddock (I have issues trying to run Haddock, so I can't check):

```haskell
newtype Final m z a where
  WithWeaving :: (  forall f
                  . Functor f
                 => f ()
                    -- ^ Initial effectful state. Blablabla...
                 -> (forall x. f (z x) -> m (f x))
                    -- ^ Conversion with effectful state to the final monad. Blablabla...
                 -> (forall x. f x -> Maybe x)
                    -- ^ Inspector. Blablabla...
                 -> m (f a)
                 )
              -> Final m z a
```

src/Polysemy/Final/Async.hs
KingoftheHomeless: The only thing I'm worried about is if `-Final` interpreters somehow perform worse. (I don't see how, but I haven't checked). Otherwise, yeah, `lowerAsync`, `lowerError`, `lowerResource`, and `runFixpointM` should all be safe to deprecate.

src/Polysemy/Final.hs
KingoftheHomeless: For each action in `Final`, so this inherits the work duplication issue of traditional `lower-` interpreters.

src/Polysemy/Final/Internal.hs
isovector: The goal is for documentation. Unfortunately haddock is too dumb to annotate higher rank functions; but I'd be happy with comments here if you feel the GADT is too heavyweight.

src/Polysemy/Final.hs
KingoftheHomeless: What I require here isn't an isomorphism, but I digress: an example would be between `m` and `Codensity m`. `lowerFinal` is also interpreted in terms of `finalToFinal` (but it doesn't need to be).

src/Polysemy/Final.hs
KingoftheHomeless: Yeah, as long as you're only working with polymorphic constraints on the effect list, you can use `runFinalPure` any time you could use `run`.

src/Polysemy/Fresh.hs
isovector: I don't have a strong preference. But I use this effect a lot for generating user-facing symbols, which are definitely nicer if they're monotonic. I don't think the ambiguity will be an issue for any user-facing code that uses `Fresh`.

src/Polysemy/Final.hs
isovector: Fair enough, thanks!

src/Polysemy/Final.hs
KingoftheHomeless: What I mean by this is that `interpretFinal` does not let you use any actions of `Sem r` in any way, shape, or form as you interpret each action. For example, you can't read an `Input` for each action, which you could do using `interpretH` + `withStrategic`:

```haskell
ex = interpretH $ \case
  Foo im -> do
    i <- input
    m <- ex (runT (im i)) 
    raise $ withStrategic $ do
      m' <- runS m
      ...
```

KingoftheHomeless:
```Haskell
Member (Final m) r => ((forall x. Sem r x -> m x) -> m a) -> Sem r a
```
Is not possible using `Final`; you have to use the inspector or distribution.
It's in this sense the power `Final` provides is slightly weaker than a straight-up natural transformation.

src/Polysemy/Cont/Internal.hs
KingoftheHomeless: I mostly built these handlers by kludging the free variables together until they type-checked, without any greater regard if the way I constructed them was right. It seems to work, but since I constructed them haphazardly, they're dubious.

src/Polysemy/Cont/Internal.hs
KingoftheHomeless: The `Any` corresponds to `a` when the `ViaFreshRef` was originally created (as opposed to a `contramap`ped `ViaFreshRef`). We need it because we can't scope over that type variable, because then we wouldn't be able to lower other effects through. This is also why `uniq` is so important here; we create an `uniq` when we create the original `ViaFreshRef`, and then when we receive a backtrack exception, we check if it's `uniq` matches the one we created. If they do, we know the `Any` is the original `a`, and is thus safe to `unsafeCoerce` to `a`.
Existentializing doesn't help here, because `Subst` places no constraints (like `Typeable`) upon the original `a`, so we'd have to resort to `unsafeCoerce` anyway.

src/Polysemy/Final/Internal.hs
KingoftheHomeless: I mean, I kinda feel that way? I like the mantra of `withWeaving $ \s wv ins -> ...` once you understand how it works.

Besides, even though we won't be able to annotate the constructor, we will be able to annotate the derived `withWeaving` action.

By the way, something that just cropped up in my mind; should `WithWeaving` be renamed to `WithWeavingToFinal`? More verbose, but also mirrors `withLowerToIO`.

src/Polysemy/Final/MTL.hs
KingoftheHomeless: [At the very least, this call site.](https://github.com/KingoftheHomeless/polysemy-zoo/blob/6527992539da9a8ff2497ede3af7dd0fa2d4e691/test/FinalSpec.hs#L112)

src/Polysemy/Fresh.hs
KingoftheHomeless: Perhaps what we could do is have `runFreshUnsafe` use a newtype of `Integer` instead (just like `Unique`), so that can be used to provide a generic user-facing symbol, and have `runFreshNumUnsafe` seperate.

src/Polysemy/Final/Internal.hs
KingoftheHomeless: Sure thing.

src/Polysemy/Final/Writer.hs
```diff
@@ -0,0 +1,114 @@
...
+  s <- readTVar tvar
+  writeTVar tvar $! s <> o
+{-# INLINE runWriterTVar #-}
+
+-- TODO(KingoftheHomeless): Make this mess more palatable
+--
+-- 'interpretFinal' is too weak for our purposes, so we
+-- use 'interpretH' + 'withWeaving'.
+runWriterTVarAction :: forall o r a
+                          . (Member (Final IO) r, Monoid o)
```
KingoftheHomeless: This could be generalized to `Member (Final m) r, MonadIO m, MonadMask m`, but I doubt that's worth it.

src/Polysemy/Fresh.hs
KingoftheHomeless: Yeah, I like that idea. I'll implement that.

src/Polysemy/Final.hs
KingoftheHomeless: Then this is how `runFinal` will look.

src/Polysemy/Cont/Internal.hs
isovector: that's how i built the whole library ;)

src/Polysemy/Final.hs
isovector: Dang; was really hoping we'd get that by knowing the final monad.

src/Polysemy/Final.hs
KingoftheHomeless: Actually, it shouldn't, should it? Seeing how `runFinalPure`'s `forall x. Union r (Sem r x) -> Identity x` shouldn't be invoked, just like absurdU in `run`.

With that in mind, yeah, you could replace `run` with `runFinalPure` if you want. However, `runFinalPure` really only exists for use together with `fixpointToFinal`.

isovector: I'd prefer to have an `eliminate :: Sem (Final Identity ': r) a -> Sem r a` than this special case

src/Polysemy/Final.hs
isovector: In that case I think I'm confused. Can I write this?

```haskell
runStateIO :: Member (Final IO) r => s -> Sem (State s ': r) a -> Sem r a
runStateIO s m = do
  ref <- final $ newIORef s
  runStateInIORef ref m
```

and if so, will it allocate the ioref exactly once?

src/Polysemy/Final/Async.hs
isovector: They're questionable enough that we should just deprecate them and deal with perf if it comes to that.

src/Polysemy/Final/Internal.hs
isovector: Both sound good to me.

src/Polysemy/Final/MTL.hs
isovector: remarkable.

src/Polysemy/Final/Writer.hs
isovector: Let's keep the mtl stuff as far away as possible :)

src/Polysemy/Final.hs
KingoftheHomeless: That will allocate the `IORef` exactly once _unless_ you use `lowerFinal`, and you use `runStateIO` in the argument of `lowerFinal`.

So this will allocate the `IORef` exactly once:
```
runFinal . runStateIO s . ...
```

But say you instead do this:
```
(runFinal . embedToFinal . runStateIO s .@ lowerFinal) . ...
``` 
Then _this_ will allocate the `IORef` multiple times.

src/Polysemy/Final/Async.hs
KingoftheHomeless: Ok, I'll update the comments to say "prefer `-Final` interpreters".

src/Polysemy/Fresh.hs
KingoftheHomeless: How do you feel about `runFreshEnumUnsafe` instead of `runFreshNumUnsafe`?

src/Polysemy/Final/Writer.hs
KingoftheHomeless: Sure.

src/Polysemy/Final.hs
KingoftheHomeless: You can't write `eliminate`, unfortunately; it would require a natural transformation `forall x. Sem r x -> x`.

KingoftheHomeless: We can remove `runFinalPure` if we're ok with forcing users to use `runIdentity . runFinal`.

src/Polysemy/Final.hs
KingoftheHomeless: I'll add a _Beware_ to `lowerFinal` to explain this.

src/Polysemy/Fresh.hs
```diff
@@ -0,0 +1,141 @@
...
+
+newtype UniqueRef = UniqueRef Integer
+  deriving (Eq, Ord, Hashable)
```
KingoftheHomeless: This stings a lil bit. I feel `Hashable` is very important instance to have, so unfortunately, this incurs an extra dependency.

src/Polysemy/Final.hs
isovector: Yeah, let's do that.

src/Polysemy/Final.hs
isovector: wat. Is `lowerFinal` going to go away when we gut all the other lowers and `.@`?

src/Polysemy/Final/Async.hs
isovector: Thanks!

src/Polysemy/Fresh.hs
isovector: Love it. Make it so!

src/Polysemy/Fresh.hs
isovector: Let's _just_ provide the `Unique` and `Enum` instances, and then not need the dependency?

KingoftheHomeless: I don't understand. Do you mean removing `runFreshUnsafe`?

src/Polysemy/Final.hs
KingoftheHomeless: You're right, it doesn't belong on Polysemy proper. I basically added it as a contingency.

I can only think of one extremely contrived use-case:

1. You want your final monad to be `m`.
2. You have a bunch of `Final n` constraints, where `n` is different from `m`.
3. You only have a natural transformation `n` to `m`, not `m` to `n`, so `finalToFinal` won't work.
4. BUT you're able to interpret the rest of your effect stack either in terms of `m` or `n`.

In this case, you can use `lowerFinal` and `runEmbedded` to make it all work out.

But honestly, yeah. `lowerFinal` can stay here. I also don't have much justification behind `runFinalSem`, so it can stay here too. It's mostly just neat that you can implement `lowerFinal` using it.

src/Polysemy/Final/Internal.hs
```diff
@@ -0,0 +1,230 @@
...
+--
+-- A @'Polysemy.Internal.Union.Weaving' 'Final'@ provides these components,
+-- hence the name 'ThroughWeavingToFinal'.
+type ThroughWeavingToFinal m z a =
+     forall f
+   . Functor f
+  => f ()
+  -> (forall x. f (z x) -> m (f x))
+  -> (forall x. f x -> Maybe x)
+  -> m (f a)
```
KingoftheHomeless: What do you think of this? For some reason it slipped my mind `withWeavingToFinal` can't be annotated either, so I created this type which I thoroughly document, and use it in `WithWeavingToFinal`

src/Polysemy/Fresh.hs
KingoftheHomeless: @isovector Can you clarify what you mean?

isovector: We don't need this `UniqueRef` thing, right? And if we dont have it then we don't need a dependency on hashable :)

src/Polysemy/Final/Internal.hs
isovector: love it!

src/Polysemy/Fresh.hs
KingoftheHomeless: `UniqueRef` is meant as a user-facing symbol, so users don't have to cook up their own, and don't have to deal with ambiguity error messages when using the interpreter.

It's nice to have, but you're right in that we don't strictly need it. Personally, I feel that `hashable` is such a small and universal dependency that the cost of using it here is next to nothing. It's also already a transitive dependency, so we don't exactly force users to download more packages. In the end though, it's your decision.

src/Polysemy/Final/Internal.hs
KingoftheHomeless: Good! I'll keep this then.

src/Polysemy/Final/Writer.hs
```diff
@@ -20,10 +29,53 @@ runWriterTVar tvar = runWriterTVarAction $ \o -> do
...
+-- You should always use this instead of 'writerToIOFinal' if the monoid
+-- is a list, such as 'String'.
+--
+-- /Beware/: Effects that aren't interpreted in terms of 'IO'
+-- will have local state semantics in regards to 'Writer' effects
+-- interpreted this way. See 'Final'.
+writerToIOAssocRFinal :: (Member (Final IO) r, Monoid o)
+                      => Sem (Writer o ': r) a
+                      -> Sem r (o, a)
+writerToIOAssocRFinal =
```
KingoftheHomeless: Here are some more interpreters I imagine will be useful. `writerToIOFinal` just removes the basic boilerplate usually attached to `-IORef` / `TVar` interpreters, and `writerToIOAssocRFinal` writes out the not-so-basic boilerplate needed to right-associate `tell`s for `writerToIOFinal`.

Don't worry, I won't add anything more to this pull request.

~~Unless you want a `Mask` effect~~

KingoftheHomeless:
@isovector Let's wrap up the last of the unresolved questions so we can merge this in!

src/Polysemy/Fresh.hs
KingoftheHomeless: Ok, then I'll remove it.

KingoftheHomeless:
Ready to merge!

KingoftheHomeless:
Oh, I get to do it now! :)

PR #46, opened 2019-08-30
Changes to accommodate v1.2

KingoftheHomeless:
This can be merged once we decide to release polysemy v1.2


isovector:
:heart: it, thanks! I'll look into cutting a new release tomorrow.

src/Polysemy/Final/More.hs
```diff
@@ -0,0 +1,32 @@
+module Polysemy.Final.More where
```
KingoftheHomeless: Not fond of this module name. Suggestions for better one?

isovector: Do we need these?

KingoftheHomeless: Strictly, no, but `runFinalSem`, at the very least, does express some power which can't be achieved using the existing interpreters of `Final`. I'd like to keep that one, at least, because it's plausible that it's usable.

Besides, this _is_ `polysemy-zoo`.

isovector:
Are we good to land 1.2? I'll do it today

KingoftheHomeless:
We're ready.

I'm planning on a `Tagged` effect, but it probably belongs in the zoo.

KingoftheHomeless:
Or hell, `Tagged`'s honestly already ready. This is its components:
```haskell
newtype Tagged k e m a where
  Tagged :: forall k e m a. e m a -> Tagged k e m a
```
```haskell
tag :: forall k e r a
     . Member (Tagged k e) r
    => Sem (e ': r) a
    -> Sem r a
```
```haskell
untag :: forall k e r a
       . Sem (Tagged k e ': r) a
      -> Sem (e ': r) a
```
```haskell
retag :: forall k1 k2 e r a
       . Member (Tagged k2 e) r
      => Sem (Tagged k1 e ': r) a
      -> Sem r a
```

Do you think it should be part of `polysemy`?

KingoftheHomeless:
Should have said that to start it.

Well, in any case, it can wait.

isovector:
Ahh, sorry! I'm happy to push out a 1.2.1.0 sooner than later if we want to get this in.

KingoftheHomeless:
The thorny thing about `Tagged` is the question if we should create predefined `tagged-` variants for everything. That's a hairy enough question for me not to put it in right now.

PR #51, opened 2019-09-08
Fix revPut and Tagged docs

KingoftheHomeless:
`revPut` was the same as `revGet`. This mistake makes me mad, because I had written the proper definition, and even confirmed it during a double-check. Perhaps while working on `RevState` I reverted to an older version of the file for some reason.

@isovector Is this worth an v0.6.0.1 release? `revPut` is the least important of all `RevState`'s methods; you'd typically use `revModify` instead (since you can't do `revGet >>= revPut . f` without getting into an infinite loop).

### DIFF
--- a/src/Polysemy/RevState.hs
+++ b/src/Polysemy/RevState.hs
@@ -50,8 +50,9 @@ revGet = revState $ \s -> (s, s)
 -- | Sends a new state into the past.
 revPut :: forall s r
         . Member (RevState s) r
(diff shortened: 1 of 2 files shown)


isovector:
Nice catch. Yeah, i'll push out 0.6.0.1.

KingoftheHomeless:
@isovector Reminder to push out 0.6.0.1

isovector:
done

Issue #40, opened 2019-07-15
Absorbing exceptions.MonadThrow and exceptions.MonadCatch


adamConnerSax:
So, I thought this would be an easy extension of the absorbMonadError but not so much.  Or I'm missing something.
The exceptions thing is quantified (?) over the exception type.  So I'm having trouble getting all the types--the ```e``` I need to specify for Polysemy.Error vs. the ```e``` that is forall'd in exceptions.
My attempt so far is this:
https://github.com/adamConnerSax/glm-haskell/blob/master/src/Polysemy/ConstraintAbsorber/MonadCatch.hs
but that's just at a random stopping point of trying quantified constraints and then not and then throwing things and then giving up.
Anyway, ideas are welcome.  If someone can point me in the right direction, or explain why it'll never work, that would be much appreciated.

KingoftheHomeless:
Use `SomeException` as the error type. Here's a POC:

```haskell
import qualified Polysemy.Error as P

instance Member (Error SomeException) r => MonadThrow (Sem r) where
  throwM e = P.throw (toException e)

instance Member (Error SomeException) r => MonadCatch (Sem r) where
  catch m h = m `P.catch` \e -> case fromException e of
    Just e' -> h e'
    _         -> throwM e
```

adamConnerSax:
Ah!  Hadn’t tried that. Or thought if it. Thanks! Will try.

adamConnerSax:
With a little fiddling that worked. Thanks!
@isovector, would you like this added to the zoo?  I think it belongs there, since it's just part of what ```Error``` can absorb.  I can add some tests and submit a PR, though probably not until late this week or early next. I don't think it adds much in the way of dependencies but I know you like to keep things lean...
Here's the latest (obviously not in the zoo right now.  I just needed it for something else.  but it's ready to drop in): https://github.com/adamConnerSax/glm-haskell/blob/master/src/Polysemy/ConstraintAbsorber/MonadCatch.hs
Also, though it's beyond me without a lot more work, it might be interesting to see if we can absorb ```MonadMask``` which involves some sort of bracket. See ```generalBracket``` in 
https://hackage.haskell.org/package/exceptions-0.10.2/docs/Control-Monad-Catch.html.

isovector:
[...]

PR #54, opened 2019-10-07
Added PolySemy.ConstraintAbsorber.MonadCatch and tests.  

adamConnerSax:
Here's the PR to resolve https://github.com/polysemy-research/polysemy-zoo/issues/40.

- Adds absorber for ```MonadCatch``` (as well as ```MonadThrow```, but that is a superclass of ```MonadCatch``` so I just use the ```MonadCatch``` absorber for both).  
- Added a custom run-function, ```runErrorForMonadCatch```, as well, because people will usually want to map the ```SomeException``` type to something less opaque so there will almost always be a ```runError . mapError f``` sort of thing and I just made it explicit.  But maybe that's bad form since it's trivial?  The oddness of ```SomeException``` tripped me up a bit in testing so I thought I would add the run function to help others with the same issue.
- Added some tests, throwing exceptions and handling some.
- Added the exceptions package to package.yaml
- Bumped the polysemy-zoo version to 0.7.0.0 in package.yaml as well.  

Note: I didn't attempt to add an absorber for ```MonadMask``` because that gets into asynchronous exceptions, etc. and I don't know much about all that.  But that seems like a useful thing to add in the future if anyone is interested.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -31,6 +31,7 @@ dependencies:
 - random >= 1.1 && <1.2
 - reflection >= 2.1.4 && < 3.0.0
 - transformers >= 0.5.2.0 && < 0.6
+- text >= 1.1.0 && < 1.3 
 - ghc-prim >= 0.5.2 && < 0.6
 - hedis >= 0.10 && < 0.13
 - async >= 2.2 && < 3
@@ -76,5 +77,5 @@ tests:
     - hspec-discover  >= 2.0
     dependencies:
     - polysemy >= 1.2.0.0
-    - polysemy-zoo
+    - polysemy-zoo    
     - hspec >= 2.6.0 && < 3
--- a/package.yaml
+++ b/package.yaml
@@ -1,5 +1,5 @@
 name:                polysemy-zoo
-version:             0.7.0.0
+version:             0.6.0.1
 github:              "isovector/polysemy-zoo"
 license:             BSD3
 author:              "Sandy Maguire"
--- a/package.yaml
+++ b/package.yaml
@@ -1,5 +1,5 @@
 name:                polysemy-zoo
-version:             0.6.0.1
+version:             0.7.0.0
 github:              "isovector/polysemy-zoo"
 license:             BSD3
 author:              "Sandy Maguire"
@@ -78,4 +78,3 @@ tests:
     - polysemy >= 1.2.0.0
     - polysemy-zoo
     - hspec >= 2.6.0 && < 3
-    - text > 1.2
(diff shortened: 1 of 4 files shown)


package.yaml
```diff
@@ -1,5 +1,5 @@
 name:                polysemy-zoo
-version:             0.6.0.1
+version:             0.7.0.0
```
isovector: Please leave the version as it was, I'll get confused the next time I have to release it otherwise! But thanks for the thought!

src/Polysemy/ConstraintAbsorber/MonadCatch.hs
```diff
@@ -0,0 +1,115 @@
...
+
+------------------------------------------------------------------------------
+-- | Like 'E.runError' but applies a given function from 'SomeException'
+-- to some other type, typically something less opaque.
+-- e.g.:
+--  @runErrorForMonadCatch C.displayException@
+-- 
+-- @since 0.7.0.0
+runErrorForMonadCatch
+  :: (C.SomeException -> e)
```
isovector: This can't be useful, can it? `SomeException` contains an existential and thus has infinite possibilities inside. Maybe this should be a `SomeException -> Maybe e`, which introduces the `Error e` effect into scope?

adamConnerSax: I mean, I use it in the tests since I couldn't use HSpec's ```shouldBe``` with ```SomeException``` since it doesn't have an ```Eq``` instance.  I am assuming the function you give it will almost always invoke ```fromException``` and then handle the ```Maybe``` cases.  But your version doesn't allow the very simple use of ```displayException``` which strikes me as a likely desired use-case.  So I don't know.

adamConnerSax: See https://github.com/adamConnerSax/polysemy-zoo/blob/AddMonadCatch/test/ConstraintAbsorberSpec.hs#L55 
and
https://github.com/adamConnerSax/polysemy-zoo/blob/AddMonadCatch/test/ConstraintAbsorberSpec.hs#L148

isovector: I'm not super convinced; if you require that function to be total you're just going to force `e` to have a `Nothing` equivalent, eg your `UnknownException`.

adamConnerSax: Except in the displayException case.  Should there be two run functions, one with ```Maybe``` and one for just the display case?  Let me push a version like that and see what you think...

adamConnerSax: Okay, pushed a new version with a text specific run function and one that is explicit with the ```Maybe```. 
I can see dropping these but ```SomeException``` is weird and I'd like to provide some way out of the wilderness for a user who is stuck in the weirdness.

isovector:
Excellent, thanks so much!

PR #56, opened 2019-11-12
Polysemy.Input.Streaming

isovector:
As discussed in https://github.com/polysemy-research/polysemy/pull/278

### DIFF
--- a/src/Polysemy/Input/Streaming.hs
+++ b/src/Polysemy/Input/Streaming.hs
@@ -1,0 +1,46 @@
+module Polysemy.Input.Streaming where
+
+import           Data.Functor.Of
+import           Polysemy
+import           Polysemy.Input
+import           Polysemy.State
+import qualified Streaming as S
+import qualified Streaming.Prelude as S
+
+
+runInputViaStream
+    :: S.Stream (Of i) (Sem r) ()
+    -> InterpreterFor (Input (Maybe i)) r
+runInputViaStream stream
+  = evalState (Just stream)
+  . reinterpret ( \Input ->
+      get >>= \case
+        Nothing -> pure Nothing
+        Just s ->
+          raise (S.inspect s) >>= \case
+            Left () -> pure Nothing
+            Right (i :> s') -> do
+              put $ Just s'
+              pure $ Just i
+  )
+
+
+runInputViaInfiniteStream
+    :: forall i r
+     . (forall x. S.Stream (Of i) (Sem r) x)
+    -> InterpreterFor (Input i) r
+runInputViaInfiniteStream stream
+  = evalState (stream :: S.Stream (Of i) (Sem r) ())
+  . reinterpret ( \Input -> do
+      s <- get
+      raise (S.inspect s) >>= \case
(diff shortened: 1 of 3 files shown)


src/Polysemy/Input/Streaming.hs
```diff
@@ -41,6 +43,30 @@ runInputViaInfiniteStream stream
...
+    => S.Stream (S.Of (Either i1 i2)) (Sem r) ()
+yieldRace = do
+  z <- S.lift $ withStrategicToFinal $ do
+         input1 <- runS input
+         input2 <- runS input
+         pure $ fmap sequenceEither $ A.race input1 input2
+  S.yield z
+
+
+sequenceEither :: Functor f => Either (f a) (f b) -> f (Either a b)
```
KingoftheHomeless: We don't want to expose this, do we?

isovector:
I'm very partial to `streaming` over `pipes` and `conduit` --- it's faster, has a simpler API, and has good interop in both directions.

KingoftheHomeless:
Actually, the contender I had in mind was `streamly`; although I'm not that familiar with the streaming ecosystem, so I can't say which is better.

src/Polysemy/Input/Streaming.hs
```diff
@@ -0,0 +1,85 @@
...
```
googleson78: This is `bisequence`, is it not?

isovector: It's not, unfortunately --- we only have a `Functor`, not an `Applicative`

src/Polysemy/Input/Streaming.hs
```diff
@@ -0,0 +1,85 @@
...
+import qualified Streaming as S
+import qualified Streaming.Prelude as S
+
+
+
+runInputViaStream
+    :: S.Stream (Of i) (Sem r) ()
+    -> InterpreterFor (Input (Maybe i)) r
+runInputViaStream stream
+  = evalState (Just stream)
```
FongHou: Why wrapping the state in a Just? wouldn't below have the same semantics?

```
runInputViaStream
    :: forall i r a.S.Stream (Of i) (Sem r) ()
    -> Sem (Input (Maybe i) ': r) a -> Sem r a
runInputViaStream stream
  = evalState stream
  . reinterpret ( \Input -> do
      s <- get @(S.Stream (Of i) (Sem r) ())
      raise (S.uncons s) >>= \case
        Nothing -> pure Nothing
        Just (i, s') -> do
          put s'
          pure $ Just i
  )

```

isovector: This will duplicate the actions at the end of the computation --- consider an action that only does effects and never `yield`s:

```haskell
foo :: Stream (Of i) IO a
foo = liftIO $ putStrLn "hello"
```

under the `runInputViaStream` above, this will print `hello` for every call to `input`!

FongHou: Aha, make sense. then shouldn't Left () case have 'put Nothing'?

PR #59, opened 2019-12-12
Compact Reader effect

spacekitteh:


src/Polysemy/Reader/Compact.hs
```diff
@@ -0,0 +1,33 @@
...
+     . Member (Embed IO) r
+    => i
+    -> Sem (Reader i ': r) a
+    -> Sem r a
+runReaderWithCompacted i sem = do
+    compacted <- embed (compactWithSharing i)
+    runReaderWithExistingCompacted compacted sem
+{-# INLINE runReaderWithCompacted #-}
+
+-- | Run a 'Reader' effect with a value in a compact region. Will not add 'Local' values to the existing region, but will create a new region for it.
```
KingoftheHomeless: Please format the documentation according to how functions are typically documented in polysemy. 79 characters per line, with a 
```
-----------------------------------------------------------------------------
```
line above.

Also, you should reference `local` instead of `Local`

spacekitteh: Done!

KingoftheHomeless:
Thanks!

PR #60, opened 2019-12-30
UUID generation via Fresh effect

spacekitteh:


KingoftheHomeless:
`uuid` is a very heavy dependency for such a simple interpreter that users can roll on their own. Is there a motivating reason to support `uuid`? Unless you can convince me that it's worth it, I'm unwilling to merge this in.

spacekitteh:
I honestly just used `uuid` because I couldn't be bothered copying the random generation code and using `uuid-types` (?) instead. The default RNG doesn't provide enough randomness, so it has to mess about with bit shifting and the like. 

But you're right, it is a very simple interpreter that a user can easily roll on their own - I guess I kinda viewed `polysemy-zoo` as a "batteries included" lib?

Issue #65, opened 2020-06-25
ghc 8.10 and ghc-prim bound


adamConnerSax:
Currently, the zoo won't compile with ghc 8.10.1 because of an upper bound on ghc-prim.  It might be as simple as raising the bound.  However, when I tried to check via an "allow-newer" in cabal.project, the compiler crashed with a ghc-panic while compiling  ```Polysemy.RevState```
```
<no location info>: error:
    ghc: panic! (the 'impossible' happened)
  (GHC version 8.10.1:
	mergeSATInfo
  Left:STSTSTSVSVSV, Right:STSTSTSVSCNSNS
  Call stack:
      CallStack (from HasCallStack):
        callStackDoc, called at compiler/utils/Outputable.hs:1179:37 in ghc:Outputable
        pprPanic, called at compiler/simplCore/SAT.hs:152:20 in ghc:SAT

Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
```

isovector:
:scream:

adamConnerSax:
I know this is being tracked via https://github.com/polysemy-research/polysemy/issues/356 but this "solution" is specific to the issue with RevState in the zoo.  I've got a branch that compiles with ghc-8.10.2, but only because I used CPP to remove the ```runXXX``` function in RevState.   If it's any help debugging, the bug only triggers if the functions are exported.  They compile fine.  But if either run function is on the export list, we get the GHC panic.

For now it suits my purposes, since I can use that and start working with 8.10.  Should I make a PR?  It's hardly a real fix, and it's obviously breaking, but it's also better in some ways than not being able to use the lib at all with ghc-8.10.

KingoftheHomeless:
If it's plugin-specific, it's better to rewrite `polysemy-zoo` to not make use of it. That's a horrible amount of work, however.
I'll get to it.

adamConnerSax:
Okay!  I'l just use my fork for now, then.  Thanks!

KingoftheHomeless:
Stop-gap for this has been implemented through #66. Not closing this since the underlying bug hasn't been fixed.

spacekitteh:
Could a 0.7.0.1 release be pushed in the meantime?

KingoftheHomeless:
Done.

spacekitteh:
This needs to occur again for 9.0

PR #66, opened 2020-09-27
Don't use polysemy-plugin, remove IdempotentLoweringSpec

KingoftheHomeless:
Stop-gap for https://github.com/polysemy-research/polysemy-zoo/issues/65.
This does not address the underlying bug, so I won't close that issue.
This wasn't as much work as I thought it would be.


KingoftheHomeless:
@adamConnerSax Can you confirm if this works? I personally have issues trying to get 8.10 working with stack.

adamConnerSax:
Looks like it.  I had to do an "allow-newer: ghc-prim" for my case.  Also, I wanted to do "allow-newer: random"--because the newer random is a major performance upgrade--but then I got a compilation error:
```haskell

adamConnerSax:
So maybe, while you're there, it's worth bumping all the bounds, or at least all but random, though if that's easy to fix, it might be worth it...

adamConnerSax:
Anyway, thanks for working on it!

KingoftheHomeless:
Does that do it?

KingoftheHomeless:
Forgot the tests. Those are significantly more hellish to rewrite, so I'm going with enabling the plugin for the tests only.
Also, for some reason Travis now trips up on building `IdempotentLoweringSpec`, compared to in February. I have no idea why. Considering Idempotent Lowering has long since been de-facto deprecated in favour of `Final`, and `ImpredicativeTypes` has unspecified behaviour, I'll simply be removing that test, and not lose any sleep over it.

@adamConnerSax I'll merge this when you confirm this works.

adamConnerSax:
I’ll test tomorrow AM!

adamConnerSax:
All works on my end, thanks!

PR #67, opened 2020-10-22
Add runKVStoreAsKVStore

locallycompact:
Hi, I wonder if I could add this as an experiment. Depends on optics.


locallycompact:
I'm in two minds as to whether k v k' v' or k k' v v' is more natural.

PR #72, opened 2021-07-28
add polysemy-readline to list of integrations

lehmacdj:
I've just released `polysemy-readline` version `0.2.0.0` on hackage and am happy enough with it to want to advertise it publicly here. It provides an effect that offers compatibility with the `haskeline` package.


tek:
We should really improve this list, there are so many libraries around

tek:
and move it to the main repo

Issue #74, opened 2022-03-07
IdempotentLowering and ghc 9.2


adamConnerSax:
polysemy-zoo (hackage latest and d1564f04a2089472a22104f9ecce94674a56ef77) fails to compile on ghc9.2.2 because of errors in IdempotentLowering.hs. The first:
```haskell
src/Polysemy/IdempotentLowering.hs:26:16: error:
    • Couldn't match representation of type: forall x1. m1 x1 -> n1 x1
                               with that of: m x -> n x
        arising from a use of ‘coerce’
    • In the first argument of ‘($)’, namely ‘coerce’
      In the second argument of ‘($)’, namely ‘coerce $ Nat f’
      In the expression: pure $ coerce $ Nat f
    • Relevant bindings include
        f :: forall (x :: k). m x -> n x
          (bound at src/Polysemy/IdempotentLowering.hs:26:5)
        nat :: (forall (x :: k). m x -> n x)
               -> base (forall (x :: k). m x -> n x)
          (bound at src/Polysemy/IdempotentLowering.hs:26:1)
   |
26 | nat f = pure $ coerce $ Nat f
   |                ^^^^^^
```

I took a shot at figuring this out, googled a bit. etc. but I got nowhere.

KingoftheHomeless:
`IdempotentLowering` is old, unreliable, outclassed by `Final`, and deprecated. I suggest removing the module altogether.
This makes for a good first issue!

ShapeOfMatter:
To clarify; this issue is resolved, yes?

KingoftheHomeless:
Yup; merged and released.

PR #75, opened 2022-03-07
Removed IdempotentLowering (for ghc 9.2.2)

adamConnerSax:


adamConnerSax:
resolves (https://github.com/polysemy-research/polysemy-zoo/issues/74)

KingoftheHomeless (changes requested):
This needs to be added to the release notes, under unreleased changes. This constitutes a breaking change.

ChangeLog.md
```diff
@@ -1,4 +1,7 @@
 # Changelog for polysemy-zoo
+## 0.8.0.0
+- Removed IdempotentLowering since “IdempotentLowering is old, unreliable, outclassed by Final, and deprecated.”
```
KingoftheHomeless, line 3: Have this under a `Breaking Changes` sub-sub-section (see other releases in the changelog), and put it under `Unreleased Changes` instead of 0.8.0.0. `Unreleased Changes` is currently at the bottom of this file, but we've learned the hard way in `polysemy` that's bad -- please move it to the top, where you currently have 0.8.0.0.

As for the message, I suggest the following:
```
- Removed `Polysemy.IdempotentLowering`, previously deprecated in favor of `Polysemy.Final`. 
```

KingoftheHomeless (changes requested):

polysemy-zoo.cabal
```diff
@@ -7,7 +7,7 @@ cabal-version: 1.12
 -- hash: 21ba8ceb18ae6184f65237f20847dd9e3f23ad2838d46fc1cd6af719dac66eb3
 
 name:           polysemy-zoo
-version:        0.7.0.2
+version:        0.8.0.0
```
KingoftheHomeless, line 10: This version bump doesn't have a corresponding change in the `package.yaml`. I don't see any reason to bump the version quite yet, so if you've changed the `package.yaml` locally, please return the version number to 0.7.0.2, rerun `stack build`, and push up the `.cabal` field (properly building it matters for the hash).
Alternatively, if you don't use `stack`, I could do it.

KingoftheHomeless (changes requested):

polysemy-zoo.cabal
KingoftheHomeless: Actually, thinking about it, 9.2 compatibility is important enough to release this change ASAP, so I was thinking to cut a release tomorrow. So your choice -- either bump the version number in `package.yaml` or not. Either way, `stack build` will need to be run for the hash (whether it be you or I who does it).

adamConnerSax: I don't use stack and I'm not quite sure what you're asking here...

KingoftheHomeless: I'll do it then, tomorrow. Now I'm going to sleep.

KingoftheHomeless (approved):
Updated the cabal file. I forgot that stack no longer generates a hash with its cabal files, nowadays.

KingoftheHomeless:
Thanks for this!

PR #79, opened 2022-09-14
Changes for GHC 9.2

spacekitteh:


tek:
I updated `package.yaml` and merged locally!

PR #80, opened 2022-11-21
Use Sequence instead of List in Floodgate

spacekitteh:
So that we can `snoc` rather than consing then reversing. This does, however, introduce a limit to the number of actions that can be held (specifically, `maxbound::Int`.)


tek:
seems reasonable, but I haven't ever used this. anyone else have opinions?

tek:
right. strange

isovector:
yeah this is suss. reading through it it should definitely reset the state on `Release`.

